PR #371, opened 2013-11-09
Use Font Awesome Icons

jimolina:
If you want to use Font Awesome Icon instead of default glyphicon bootstrap:

bootstrap-select.js

In the SELECT LIST element with the data-icon attribute:

that: data-icon="fa-heart"

 instead of

_this: data-icon="icon-heart"


yokas:
Does Font awesome still not be supported?

caseyjhol:
It is supported - set `iconBase: 'fa'`.

yokas:
Great! thanks, much better if it is stated in README

PR #479, opened 2014-02-15
Labels with close buttons for multi-selects.

cha55son:
This plugin is great but I felt the multi-select could be improved from a usability point of view. The labels are 'opt-in' via the `'labels': true` property. Refer to the last select on `test.html` for a demo.
###### Issues:
- Limited width of the select box causes many selected options to be hidden.
- Readability of comma delimited list of strings.
- There is no way to remove selected options without opening the select box.
###### Improvements (Only for multi-selects):
- Labels are used instead of a comma delimited list of strings.
- Labels have a click-able 'unselect' button. Users can unselect options without opening the select box.
- The select box grows as more options are selected.

### DIFF
--- a/bootstrap-select.css
+++ b/bootstrap-select.css
@@ -6,6 +6,44 @@
  * Licensed under the MIT license
  */
 
+.label.bootstrap-select-label {
+    position: relative;
+    top: -2px;
+    margin-right: 5px;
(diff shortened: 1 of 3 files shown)


NicolasCARPi:

PR #489, opened 2014-02-25
Fixed size:auto in modals

tavicu:
In modals size:auto change the modal body height.
modal-issue


caseyjhol:
This appears to no longer be an issue.

PR #519, opened 2014-03-14
added the new option numberOfStartOpts

beligante:
This new option turns you capable to load only the number of options
that you pass as argument, and show the others options when the user
scrolls the list of the options. That makes the bind of the element more
fast when you have many options in a select. 

Here is a sample: http://jsfiddle.net/J83c9/


caseyjhol:
No longer needed with the new `virtualScroll` option.

PR #602, opened 2014-05-20
added option for custom dropdown caret html

finnfiddle:
As the title says I added the ability to specify the caret html as an option so that it can be replaced with something other than `<span class="caret"><span>`. I did this so that it can be used with other icon fonts.
See my working example: .

``` javascript
$('.selectpicker').selectpicker({
    caretHTML: '<span class="entypo-down-open-big"></span>'
});
```


heyallan:
+1

busla:
+1

rogelio2k:
+1

ccdiego5:
not working for me :(

PR #636, opened 2014-07-07
wrap by '.input-group-addon' support.

icai:
the html template like this:

```
<div class="form-group">
    <div class="col-md-10 col-sm-10">
        <div class="input-group">
            <span class="input-group-addon input-group-addon-bootstrap-select">
                <select name="" id="" class="selectpicker ">
                    <option value="" data-content=''></option>
                </select>
            </span>
            <input type="text" placeholder="" class="form-control" >
        </div>
    </div>
</div>
```

### DIFF
--- a/bootstrap-select.css
+++ b/bootstrap-select.css
@@ -311,3 +311,17 @@
 .bootstrap-select .bs-actionsbox .btn-group button {
     width: 50%;
 }
+
+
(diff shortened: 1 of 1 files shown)


Eagle3386:
Then take me as one of those people. Support for ".input-group-addon" would be greatly appreciated! :+1:

PR #655, opened 2014-07-17
adding placeholder on input search

danielmariz:
now you can add a placeholder for search input
Example:
`<select class="selectpicker" data-live-search="true" data-live-search-title="Search"></select>`


anyway111:
very good, you are pro

fatihhayri:
Good property. Add documentaion this property

Issue #743, opened 2014-09-22
"selectOffsetBot" variable is computed incorrectly


GitAngel:
**selectOffsetBot** variable is being calculated like this:
>selectOffsetTop = that.$newElement.offset().top - $window.scrollTop();
>selectOffsetBot = $window.height() - selectOffsetTop - selectHeight;
The problem here is that ` $window.height()` gives the height of the full window (with scroll) and `selectOffsetTop` gives only visible part (without scroll).
That's why the value of `selectOffsetBot` will be calculated incorrectly what will produce incorrect work of `.dropup` class.

In my situation I had dropdown absolutely on the bottom of the page and I had these values as follow:
>selectOffsetTop = 840
>selectOffsetBot = 1240
Thought `selectOffsetBot` should have been `0`.

The fix should be like this:
> selectOffsetBot = $window.height() - that.$newElement.offset().top - selectHeight;

caseyjhol:
Fixed in v1.7.0-rc1 (commit 541d9a0).

PR #744, opened 2014-09-22
Fixed computation of selectOffsetBot variable

dtolstyi:
This is a fix for the issue [https://github.com/silviomoreto/bootstrap-select/issues/743]


caseyjhol:
Fixed in v1.7.0-rc1.

PR #756, opened 2014-10-10
Adds 'form-control' class to the dropdown-toggle button

cizar:
I need to highlight the control in a validation. Bootstap do it if the control has the "form-control" class.

```
<div class="form-group has-error">
  <label class="control-label" for="country">Country</label>
  <select id="country" name="country" class="form-control selectpicker">
    <option>Argentina</option>
    <option>United State</option>
    <option>Mexico</option>
  </select>
  <p class="help-block">No service available in the selected country</p>
</div>
```

has-error

Maybe there are many other solution, like the chance to define custom classes in the initialization as an option. Or use the same classes of the select element. But, since this is a Bootstrap aware plugin and the control is going to be a part of a form, I think that this solution is simple and could be included.


caseyjhol:
Permanently adding the `form-control` class can cause certain issues, as `.form-control` sets height, box-shadow, and other styles. A better alternative would be define `style: 'btn-default form-control'` or `data-style="btn-default form-control"`. Alternatively, bootstrap-select CSS now automatically styles the selectpicker according to Bootstrap defaults.

```CSS
.has-error .bootstrap-select .dropdown-toggle,
.error .bootstrap-select .dropdown-toggle {
  border-color: #b94a48;
}
```

PR #803, opened 2014-11-20
Fixed dropdown selection using space bar

illudens:
Due to a timing issue in Firefox (tested v33.1) and IE (tested v10) when a user attempts to select a dropdown item using the space bar the dropdown menu briefly closes but then reopens itself.  This works correctly in Chrome.  The fix for this was to ensure that the button is set to focus after the other operations are done on the dropdown menu.

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -750,7 +750,9 @@
           }
 
           if (!that.multiple) {
-            that.$button.focus();
+              setTimeout(function () { // BUG FIX: Fixes spacebar selection of dropdown items in FF & IE
+                  that.$button.focus();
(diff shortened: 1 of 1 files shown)


kestelle:
Here is a fiddle demonstrating the issue described by illudens which is reproducible in v1.6.4.

http://jsfiddle.net/mfrup5bL/12/

pseabury:
Above Fiddle has the same broken behavior for me in IE11 on Win8.  Chrome works as expected.

caseyjhol:
Wrapping it in a setTimeout works when pressing the spacebar initially, but the problem persists for future presses.

PR #845, opened 2014-12-31
Fix bug .selectpicker("val", "xxx") not work in ie9

xiaoyaojian:
Fix that bug,this is some different processs for arguments in IE9,I
build a array,and push the item of arguments into my array one by
one,and it can work well


fredw:
+1

t0xicCode:
Sounds good. I'll merge in a bit.

t0xicCode:
I merged #868 does that one fix it here too?

PR #847, opened 2015-01-04
Allow creation of dropdown items on the fly

anush:
This feature allows creating new dropdown items on the fly. This is useful when you want to provide a set of pre-defined choices in the dropdown, but also want to allow the user to enter _other_ values.

This feature piggy-backs off the existing _live search_ functionality pretty heavily. When enabled, instead of indicating "no results", the search will instead give you the option to create an item matching the search text. Note that this feature only works if live search is enabled. Without live search, the usability is not great, so left them tied together.
### New options

| Name | Type | Default | Description |
| --- | --- | --- | --- |
| `createItem` | boolean | false | Allows creation of new dropdown items on the fly |
| `createItemText` | string | "Create: {0}" | Let's you configure the text shown when prompting the user to create the new item. `{0}` is where the search text will appear. |
### Usage

This feature may be enabled two ways:
1. Include `data-create-item="true"` in your `select` element
2. **or**, when instantiating the Selectpicker `$('.selectpicker').selectpicker({createItem: true});`
### Limitations

The new item is always appended to the end of the list. This means that you're not going to have the best experience when working with:
1. sorted lists (sort order will not be maintained)
2. list groups (you cannot add the new item to a specific group)

These are reasonable limitations for now. This feature can be expanded upon later to support these and any other use cases.
### Test Case

Enabled this feature for the "Lunch" dropdown in `test.html`


README.md
```diff
@@ -1,69 +1,69 @@
...
-
-* [Bug reports](CONTRIBUTING.md#bug-reports)
-* [Feature requests](CONTRIBUTING.md#feature-requests)
-
-## Copyright and license
-
-Copyright (C) 2013-2015 bootstrap-select
-
-Licensed under [the MIT license](LICENSE).
+bootstrap-select
```
anushr: Looks like I inadvertently changed the line endings changed from `\r\n` to `\n`. No changes were really made to this README.

dist/css/bootstrap-select.css
```diff
@@ -1,12 +1,11 @@
...
-
+/*!
+ * Bootstrap-select v1.6.3 (http://silviomoreto.github.io/bootstrap-select)
+ *
+ * Copyright 2013-2015 bootstrap-select
+ * Licensed under MIT (https://github.com/silviomoreto/bootstrap-select/blob/master/LICENSE)
+ */
+
 .bootstrap-select {
-  /*width: 220px\9; IE8 and below*/
```
anushr: No changes were actually made to the source CSS as part of this PR. These changes appeared after I ran `grunt dist`. Perhaps `grunt dist` was not run for the latest merge into master?

README.md
t0xicCode: I'm more surprised that the line endings were not already \n. I'll run
dos2unix everywhere.

raphjutras:
This is EXACTLY what I was looking for. I'm currently using the plugin and while in test with real users this is exactly what they need because I always let them enter "custom" entry that will not be binded to a real item in database (associated with an ID). The users tryed to enter a custom value and press "enter" and instead it "post" the form. Is this enhancement completed & stable? Thank-you very much to the contributors anushr and t0xicCode!

vin86:
Hi, i have the same problem @raphjutras . I am not able to make it work properly. 

Any tips?

Thanks

raphjutras:
I didn't find a solution so I didn't use the plugin for this reason... Tell me if you find something becaus eI just kept using my old script that doesn,t create items on the fly like I wanted to...

gbaumgart:
same here, cant simply edit the combo-box value.

borex:
Managed to get it working.
Change this part:
`
var active = that.$menu.find('.active');
if(active.hasClass('no-results') && that.options.createItem) {
`

with

`
if(that.options.createItem && that.$menuInner.find('.no-results').length == 1) {
`

NicolasCARPi:

PR #854, opened 2015-01-09
Fix keycode range that triggers opening

trask:
Maybe nitpicky, but function keys (e.g. F5, F6, ...) open the select (when it has focus) because the keycode range detection is incorrect.  This is reproducible on all the examples at http://silviomoreto.github.io/bootstrap-select/.

Attached is one solution.  Another solution is to use keypress instead of keydown.  If you prefer that solution I'm happy to modify PR.

Thanks.

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -1238,7 +1238,7 @@
 
       isActive = that.$menu.parent().hasClass('open');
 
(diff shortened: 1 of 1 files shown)


t0xicCode:
Looks good. Would you mind rebasing your PR?

trask:
Done, thanks

t0xicCode:
Thanks. I'll review and :ship:

PR #857, opened 2015-01-11
Added "Done" button for multiple select to improve usability

petrdvorak:
In case of multiple select, clicking an item in the item list cannot close the dropdown. Closing can be achieved by either clicking the select itself, or by clicking outside the area of dropdown. Some people find it difficult to discover, or simply uncomfortable since there is no visible "call to action" to finish the item selection process.

In order to fix this issue, I have introduced two new options (doneButton, doneButtonText) that - when enabled - display a "Done" button at the bottom of the select. Paddings of the element were fixed accordingly.

screen shot 2015-01-11 at 1 45 28 am

screen shot 2015-01-11 at 1 45 43 am

I don't know how to localize the default "Done" text to so many languages, though.


Khrysller:
Great!

Khrysller:
How can I use this?

petr-inmite:
@Khrysller Hello, I am sorry for a bit lacking docs on this new feature - the docs are outside github... I basically added two new options to bootstrap-select component:

- "doneButton": true / false - Indicates if the multiple select should display done button. The option works only on multiple selects (see "multiple" attribute).
- "doneButtonText": 'Done' (default) - Specifies the text on the done button - for example you can set it to custom string like "I am done with this".

You can use it like this:

```html
 <select class="multiselect" multiple>
    <option>Mustard</option>
    <option>Ketchup</option>
    <option>Relish</option>
</select>
```

```javascript
$('.multiselect').selectpicker({
    doneButton: true,
    doneButtonText: "Finito"
});
```

The change is not so huge from the source code perspective. You can have look at revision 827b47824f2c899029928a250ccfe5735da7b978 - it won't be hard to find all implementation details (the code of this component is reasonably readable, that's why I managed to contribute:)).

Please note the feature probably didn't make it to the new distribution yet (you have to use trunk). Hope this helps, if there are still some questions, please do not hesitate to ask...

t0xicCode:
It's in the 2.0.0 preview tag, but the 2.0.0 is not ready for release yet.

Khrysller:
@petr-inmite, Thanks. I am so exited to use this.
@t0xicCode, thanks. I am counting the days : )

Khrysller:
Hi. Is this in the recent release? If yes, how can I use it inside the `select` tag?

Khrysller:
Hi. Is this in the 1.6.4 release? If yes, how can I use it? Tried to add the class `multiselect` but didn't work. Thanks.

petr-inmite:
Quick and maybe a bit rude response: 

- Usage is described above.
- No, it is not in 1.6... because it will be in 2.0, as written above.
- sources are available, I linked a patch you can apply, also above...

Khrysller:
Ok, so I tried in 1.6.4 adding `data-done-button="true"` and it added a Close button. Is this a feature for 1.6.4 or before?

Side effect: The Close button triggers the submit, instead of just hidde the combobox.

Bobobrien92:
@Khrysller  I was able to stop the close button triggering the submit by editing bootstrap-select.js.  Just look for the line in the code that creates the button and add type="button" to the button html.  This prevents the submit trigger.

Khrysller:
Thanks @Bobobrien92. I think this should be at sight for @caseyjhol and @t0xicCode.

Issue #828, opened 2014-12-14
No way to specify what text to search with liveSearch option


gtristan:
I'm using the selectpicker with data-content to specify an icon and some text to be displayed for each option, this works very well.

Additionally, I'm setting the 'value' of each as a stringified number - it's very important that the value itself be a number (as it is actually a UID in a collection of data records of sorts).

I've looked at the code but cannot find any alternative text for each <option> which the selectpicker considers for the matching algorithm, this would be very helpfull.

If it's there and I haven't seen it, then - it would be nice to add this to the documentation here:
   http://silviomoreto.github.io/bootstrap-select/

Best Regards.

gtristan:
This is perfect by the way, I'll be making use of this very soon thanks.

PR #858, opened 2015-01-12
Add 'data-tokens' to specify text used in liveSearch (milestone #828)

ianmustafa:
This will fix issue #828 which currently we couldn't specify search options, or _tokens_, to each `<options>`. One can use `data-tokens` in each options to specify what text to be used as tokens for that `<option>`. If `data-tokens` is exist, it will tell `icontains()` function to use `data-tokens` as haystack instead of `<option>`'s text.


t0xicCode:
Mind rebasing?

PR #860, opened 2015-01-13
Solved Tab issue with next control and Adding RTL Support

SamerX:
Steps to reproduce issue:
- in options selectOnTab option is set to true
- press tab to get focus on the selectpicker control
- arrow keys to move up and down the options in selectpicker
- press tab key to select option
- issue: focus is still on the selectpicker control (not the next html input in DOM tree)
- same issue apply when liveSearch option is set to true

This pull request solves the problem in the two cases.

### DIFF
--- a/dist/css/bootstrap-select.css
+++ b/dist/css/bootstrap-select.css
@@ -254,4 +254,16 @@
   height: 100% !important;
   opacity: 0;
 }
+
+/*bootstrap-select RTL support*/
+body[dir='rtl'] ul.selectpicker li {
+    text-align: right;
+}
+body[dir='rtl'] button.selectpicker span {
+    text-align: right !important;
+}
+body[dir='rtl'] .bootstrap-select.btn-group .btn .caret {
+    right: auto;
+    left: 12px;
+}
 /*# sourceMappingURL=bootstrap-select.css.map */
--- a/dist/js/Tabit.js
+++ b/dist/js/Tabit.js
@@ -1,0 +1,38 @@
+/*jQuery plugin that simulates clicking Tabit
+ * Authoer : Samer Abu Rabie =>samerx@outlook.com
+ * v 1.0.0
+ */
+function Tabit() {
+    var focusable = ":input, a[href]";
+    var $from = $(document.activeElement);
+    var offset = +1;
+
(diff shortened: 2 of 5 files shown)


cuplizian:
+1! Will improve usability and UX. Great job @SamerX!

SamerX:
The issue seems to be bigger than I thought!
The solution I provided looks to find the next input control using jquery next() method, this method does get the next element of a specified element with the same container, so if the next element is in another div that doesnt share the same parent container (e.g div) then this will not work!

I managed to solve it in a more general way that simulates pressing the Tab key instead of relying on jQuery next method, I will be adding a pull request for this issue after submitting this comment.

sw90248:
Hi All, 
I'm a newb to jquery, and plugins to jquery.. 

I'm currently using the latest version : Selectpicker.VERSION = '1.11.2';

and I've tried many different ways to "turn-on" selectOnTab. 
Method 1: 
i read that you only append "data" to the options if its data-related : 
So i tried : 
<select select-on-tab="true" data-live-search="true" id="<%=fieldname%>" name="<%=fieldname%>" class="auto-completeUSStates form-control selectpicker<%if eval(fieldname_error & "_errorstart")<> empty then%>error-input<%end if%>">
Results : This didn't work. 

Then after rereading examples, I tried to embed it within the class field (coz show-tick example was using this method) : 
(per eg: http://silviomoreto.github.io/bootstrap-select/examples/) show-tick 
<select class="selectpicker show-tick">

My implementation : 
<select data-live-search="true" id="<%=fieldname%>" name="<%=fieldname%>" class="auto-completeUSStates form-control selectpicker select-on-tab<%if eval(fieldname_error & "_errorstart")<> empty then%>error-input<%end if%>">

Results: This didn't work. 

Questions:
1) Could you help guide me to whats an appropriate implementation ? 

2) And how does one determine whether or not an option should be "embedded into the class attribute"( eg show-tick)  
or if its stand-alone as an attribute (eg data-live-search="true") ?

My expectations after this option is set to true, will help user : 
   while the user is in this select tag 
   they can type letters 
   once they find a matching filtered results
   they can hit tab to select this matched result
   and another tab will take them to the next field.

NicolasCARPi:

PR #863, opened 2015-01-14
init value

arifemre:
```
    <select class="selectpicker" value="2">
      <option value="1">cow</option>
      <option value="2">bull</option>
      <option value="3">ox</option>
    </select>
```

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -174,7 +174,11 @@
           that.$button.focus();
         });
       }
-
+	    
+	    if (this.$element.attr('value') != undefined && !this.$element.find('option').is("[selected]")){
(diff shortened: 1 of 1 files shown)


t0xicCode:
Are there any advantages to using this rather than using the [selected](http://www.w3.org/TR/html5/forms.html#attr-option-selected) attribute of the `<option>`?

arifemre:
Easier to use in database operations like this

t0xicCode:
I've decided to mark it as wontfix because the `value` attribute is not part of the html spec, and the `selected` attribute works (it may require a bit more processing server-side).

I'd be willing to revisit the decision if enough people show support for the change.

PR #897, opened 2015-01-29
Stopped 'change' event from bubbling

marcjae:


t0xicCode:
Do you have a specific issue to prevent the change bubbling?

marcjae:
A perfect example would be: If the select  is in a carousel (ie. Owl) that is listening for change events. On select change the event bubbles to the parent (carousel).

t0xicCode:
Sounds good.

PR #911, opened 2015-02-07
Done trigger event

Palen:
If you have a large multiselect it could be heavy to  load and you should hide the unpicked multiselect and then show it. In order to control the timeout it's good to have an event to match the time occurred.
For example: 
$('.selectpicker').on('done', function(){
        $("#loader").fadeOut("slow");
        $("#form").removeClass('hide');
    });
Instead of:
setTimeout(function() {
        $("#loader").fadeOut("slow");
        $("#form").removeClass('hide');
    }, 700);

Thanks


caseyjhol:
This has been replaced with the `loaded.bs.select` event.

Issue #835, opened 2014-12-22
IE 9 live search issue after searching non existing item


bahaa82:
Hi,
I am using bootstrap-select  and it is working fine for me except in IE 9 , when I tried to search non existing item via the live search , then deleted the text, the options list does not show up again.I tried it also in the demo page.
Thanks in advance.

bahaa82:
Can you please advise if there is work around for this issue ?

ayush15690:
I also have the same issue, Any suggestions

haseeb1431:
Any chances to have it fixed in near future or next release?

arasu-rrk:
@silviomoreto Is it possible to include this fix in next release?

PR #925, opened 2015-02-18
Fixed IE9 live searching reset bug (including IE clear button). See #835

Contagious06:


jpcofr:
@silviomoreto is this pull request going to be included officially?

caseyjhol:
Thanks for this! Opting to not include this as we're working toward dropping IE < 11 support.

PR #926, opened 2015-02-18
SelectAll/DeselectAll (Fix #721, Fix #901, fix #921)

### LINKED ISSUE #721: SelectAll and DeselectAll issues
I found that if I want to selectAll/deselectAll via program I must show the list items first otherwise these two events don't work.

And I digged into the code and I found that at line.860 and line.865, the selector for these two events is like this: 
```javascript
this.$lis.not('.divider').not('.disabled').filter('.selected').filter(':visible').find('a').click();
```

And the selector ```.filter(':visible') ``` make this issue I guess. I commented this selector and then they work well.

So why is  ```.filter(':visible')``` added here? Thanks

### LINKED ISSUE #901: Select All / Deselect All fires multiple change events on multiple select [performance]
**Problem description**

Currently, when user clicks "Select All" or "Deselect All" buttons, following code is called:

```
1052   selectAll: function () {
1053       this.findLis();
1054       this.$lis.not('.divider').not('.disabled').not('.selected').filter(':visible').find('a').click();
1055   },
1056 
1057   deselectAll: function () {
1058       this.findLis();
1059       this.$lis.not('.divider').not('.disabled').filter('.selected').filter(':visible').find('a').click();
1060   },
````

Basically, a click event is invoked multiple times and as a result, client JavaScript code receives multiple "change" events. If these change events are connected to some "slow task", such as Ajax calls to server (for example in order to fill in second select with data based on the selection in a first select), performance suffers and additional code must be written in order to better synchronize multiple Ajax calls.

**Solution proposal**

Adding a new option `aggregateChangeEvents: true/false, default = true`. If the property is set to `false`, the component keeps the current behavior (in case someone relies on it in code). If the property is set to `true` (default), only the code from the click handler is called on all items (no "virtual clicks" are invoked).

Solution affects only "multiselect" selects. However, looking at the code, this change will require some "careful surgery". I will take the issue over if required (just call "Go!" in case you cannot pull solution of the sleve:)) but will ask for a thorough code review after I submit pull request...

### LINKED ISSUE #921: BUG: 'deselectAll' doesn't work
$("select").selectpicker('deselectAll')  <= doesn't work.
Problem is: .filter(':visible') in deselectAll function.

caseyjhol:
Fixes performance issues with select/deselect all

fix #721
fix #901
fix #921


caseyjhol:
@t0xicCode Any reason this shouldn't go through?

t0xicCode:
Nothing jumps out.

:ship:

PR #956, opened 2015-03-07
Added support for locale option

twifty:
I have made some modifications to the code to allow for a `data-local` attribute and a `locale` method. This enables users to combine all locale files into a single file. This way we don't need to perform any special handling server side. If a locale is not found it will default to the built in 'en_US' or the last used locale. Also, given an array of locales, it will try to find the best, i.e. [en_GB, en_US] if neither are found it will try 'en'.

The one problem I cannot seem to work around is the BC. This approach requires the locale to be specified and the inclusion of a single locale file after loading selectpicker no longer works.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ bower_components
 
 # Dist zip
 bootstrap-select-*.zip
+/nbproject/private/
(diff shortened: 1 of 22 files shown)


NicolasCARPi:

PR #957, opened 2015-03-07
Added an expanded option

twifty:
I have added this option to make the control play nice with Symfony forms. Symfony has a form type called 'choice', a synonym for select, which has an option called expanded. When this option is given, rather than outputting a standard select control, it will render an unordered list where each option is replaced with either a checkbox or radio, depending on the multiple attribute.

This added option simply adds an 'expanded' class to the new element. It hides the button and makes the list always visible. The functionality has not been changed.

If you know Symfony, you would also know that to produce the same effect would require creating a form theme and wrapping the existing control in a div so that the css class can be applied to the new child. Now, we just need to add an extra parameter to the form builder.

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -303,7 +303,7 @@
       this.$menu.data('this', this);
       this.$newElement.data('this', this);
       if (this.options.mobile) this.mobile();
-      if (this.options.expanded) this.expanded();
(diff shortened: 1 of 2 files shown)


caseyjhol:
This doesn't really require a separate function. Just add the expanded class to your select.
```html
<select class="selectpicker expanded">
    <option></option>
</select>
```

PR #960, opened 2015-03-09
Fix liveSearchNormalize option not working.

tiagoantunesnet:
Fix two errors:
- Function 'aicontains' was calling the 'includes' function incorrectly
- Function 'generateA' was putting full HTML into the 'data-normalized-text' attribute, when it should include only the option text


caseyjhol:
The first error has been addresses in 1.7.0-rc2. I'm leaving the second "error" though, as it allows to search subtext as well as regular text.

PR #964, opened 2015-03-12
Add main declaration to package.json

dcousineau:
Added main declaration pointing to bootstrap-select entry point to allow bundling via browserify. This will enable the common [vendor browserify recipe](https://github.com/sogko/gulp-recipes/blob/master/browserify-separating-app-and-vendor-bundles/gulpfile.js).


kdeloach:
+1

caseyjhol:
The `main` declaration already exists in package.json.

PR #980, opened 2015-03-22
Clicked option passed on change plus option click event

Towanime:
Selected option will be passed as extra data to the change event function. In addition the option click event will be triggered as well in case there are listener for that option. Useful for validations.

Example: Have to disable/enable certain options depending on the selected one, either by adding a click listener to one option or using the second parameter in the change event should be easier to test and validate what other options should be disabled or enabled.

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -906,7 +906,7 @@
           }
 
           // Trigger click event on the option
-          $option.click();
+          $option.triggerHandler('click');
         }
       });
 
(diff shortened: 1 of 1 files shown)


django23:
This breaks the multiple select. With your solution, when you select an option, the dropdown closes

Towanime:
Fixed it using $option.triggerHandler("click") instead of $option.click(). I did a little example [here](http://jsfiddle.net/Towanime/rtpxwab5/2/).

Issue #737, opened 2014-09-16
Ability to select/deselect all optgroup items


jayanthigopi:
I need to clickable optgroup option, when we click on the optgroup label to select / deselect what are all items having inside the groups are selected /deselected. This is very urgent please...

eswak:
Looking forward to this functionality too.

Also, duplicate of #680, in which @t0xicCode tagged this functionality for the 1.7.0 milestone.

rkurbatov:
It would be nice and seems logical.

obarrellet:
Looking forward for that.

caseyjhol:
For anybody who truly needs this right now, replace these lines here: https://github.com/silviomoreto/bootstrap-select/blob/master/js/bootstrap-select.js#L957-L965 with

```javascript
      this.$menu.on('click', 'li.divider, li.dropdown-header', function (e) {
        e.preventDefault();
        e.stopPropagation();
        
          if (that.multiple && $(e.currentTarget).hasClass('dropdown-header')) {
              var $options = that.$element.find('option'),
                  $optgroup = $(e.currentTarget),
                  optgroupID = $optgroup.data('optgroup'),
                  maxOptionsGrp = $optgroup.data('maxOptions') || false,
                  $optgroupLis = that.$lis.filter('[data-optgroup="' + optgroupID + '"]').not('.divider, .dropdown-header, .disabled, .hidden'),
                  selectAll = $optgroupLis.filter('.selected').length !== $optgroupLis.length;

              if (!$options.eq($optgroupLis.data('originalIndex')).parent().data('maxOptions')) {
                  $optgroupLis.each(function(index) {
                      var $this = $(this);
                      $options.eq($this.data('originalIndex')).prop('selected', selectAll);
                  });
                  
                  $optgroupLis.toggleClass('selected', selectAll);
                  
                  that.render(false);
              }
          }
          
        if (that.options.liveSearch) {
          that.$searchbox.focus();
        } else {
          that.$button.focus();
        }
      });
```
I was going to implement this, but thought it would be best to first discuss how best to handle optgroups with maxOptions set. With this code, if an optgroup has maxOptions set, clicking the optgroup label doesn't do anything. Anybody have any ideas for a different way to handle this?

tomicakorac:
I agree with your suggestion. I don't see a better way to deal with the limited number of options. Maybe it's also a good idea to add some kind of notification in that case, so that users are also aware of it?

eswak:
[...]

PR #1011, opened 2015-04-30
added feature to select optgroup and toggle options under optgroup

meetmagdalene:
If a user applies an id to an optgroup and that same optgroup id as a class of the options inside that optgroup, when a user clicks the optgroup label it toggles selecting the options in that optgroup.

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -219,14 +219,22 @@
     this.init();
   };
 
-  Selectpicker.VERSION = '1.6.5';
+  Selectpicker.VERSION = '1.6.4';
 
   // part of this is duplicated in i18n/defaults-en_US.js. Make sure to update both.
   Selectpicker.DEFAULTS = {
     noneSelectedText: 'Nothing selected',
     noneResultsText: 'No results matched {0}',
     countSelectedText: function (numSelected, numTotal) {
-      return (numSelected == 1) ? "{0} item selected" : "{0} items selected";
+      if (numSelected == 1) {
+        return "{0} item selected";
+      } 
+      else if (numSelected > 0 && numSelected !== numTotal) {
+        return "{0} items selected";
+      } 
+      else if ( numSelected > 0 && numSelected == numTotal) {
+        return "Any";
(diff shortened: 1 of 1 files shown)


Khrysller:
Can I see this feature in a jsfiddle, @meetmagdalene? Thanks.

meetmagdalene:
Will try and make one up tomorrow. I've never used JSFiddle before, so, more new things this month. :)

js/bootstrap-select.js
```diff
@@ -409,11 +399,12 @@
        * @param [optgroup]
        * @returns {string}
        */
-      var generateLI = function (content, index, classes, optgroup) {
+      var generateLI = function (content, index, classes, id, optgroup) {
```
meetmagdalene: this is where you can add the optgroup label as a li id for the .dropdown-header

js/bootstrap-select.js
```diff
@@ -470,6 +455,7 @@
             optID += 1;
 
             // Get the opt group label
+            var optgroupLabel = $this.parent().attr('label').replace(/\s+/g, '-').toLowerCase();
```
meetmagdalene: Grab the label from the optgroup

js/bootstrap-select.js
```diff
@@ -479,7 +465,7 @@
               _li.push(generateLI('', null, 'divider', optID + 'div'));
             }
 
-            _li.push(generateLI(label, null, 'dropdown-header', optID));
+            _li.push(generateLI(label, null, 'dropdown-header', optgroupLabel, optID));
```
meetmagdalene: set the optgroup label as the id for this .dropdown-header

caseyjhol:
Thanks for the pull request. Unfortunately this has a few issues, namely using an old version of bootstrap-select and changing numerous default options. Thanks for contributing, though! I was working on implementing this, but ran into a wall when trying to decide how to handle optgroups with maxOptions set. See https://github.com/silviomoreto/bootstrap-select/issues/737#issuecomment-97983484 for more information.

PR #1016, opened 2015-05-04
horizontally keep in viewport

janza:
aligns menu to the right border of the element if it doesn't fit in viewport
horizontally


caseyjhol:
Hrmm...wouldn't it be better to keep it left aligned if it doesn't fit in the viewport? Then you could at least see the beginning of each option, rather than the end? This might be useful for a right-to-left implementation, though.

janza:
You are right, this is for the cases where there is space on the left of the select box, but the right part goes out of the viewport. I'll update the pull request making sure that when there's no space on the left to stay left aligned.

PR #1053, opened 2015-06-08
Added option to display different text when all options are selected.…

matt-jobson:
… (options.allSelectedText)

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -641,6 +641,11 @@
         }
       }
 
+      //If all options are selected, use the all selected text if specified.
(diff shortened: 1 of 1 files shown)


C0ZEN:
Like that ? #1183

caseyjhol:
Simply set a `countSelectedText` function instead:

```js
    countSelectedText: function (numSelected, numTotal) {
      if (numSelected == 1) {
        return "{0} item selected";
      } else if (numSelected === numTotal) {
        return "All items selected";
      } else {
        return "{0} items selected";
      }
    }

PR #1056, opened 2015-06-08
Fix for multiselect options incorrectly appearing as disabled in Firefox

creativecoder:
I had the experience where some multiselect options were appearing as disabled in firefox, even though the original option elements where not disabled. This fixes it.

It appears that Firefox's js processing does `&&` / `||` priority differently than Chrome. (I didn't check other browsers)


creativecoder:
Thanks!

PR #1062, opened 2015-06-10
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3


vsn4ik:
:+1: for merge. Also need remove https://github.com/silviomoreto/bootstrap-select/blob/d565eabc633336789d15106b6b63eba111046d12/Gruntfile.js#L180.

PR #1106, opened 2015-07-21
Added callback function sortSearchResults

marcodelpercio:
Hi,

I've implemented a new callback option named "sortSearchResults" which will let the users define a custom sorting when searching for options via live-search. 
When implemented the function passes 2 arguments:
- The current search query string
- The list of the filtered option items (i.e. <li> elements) 
  and expects to return a again an array of option items which will eventually be sorted according to whatever custom user-defined logic.

Currently when searching through live-search bootstrap-select is just "hiding" the non-matching items in the DOM therefore the sorting of the search results is exactly the same order of the items in the DOM. This might not necessarily be  the expected order of the search results.
In my case I wanted to sort the search results alphabetically and give priority to those matching options  "starting with" the current query string. 
I could not find any option to define custom sorting of the search results therefore I'm pulling a request here. See the example in test.html 

P.S. I have no idea why the lines from [8-23] are marked as changed as I haven't really edited that part

### DIFF
--- a/dist/js/bootstrap-select.js
+++ b/dist/js/bootstrap-select.js
@@ -333,7 +333,8 @@
     maxOptions: false,
     mobile: false,
     selectOnTab: false,
-    dropdownAlignRight: false
+    dropdownAlignRight: false,
+    sortSearchResults: null
   };
 
   Selectpicker.prototype = {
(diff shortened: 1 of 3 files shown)


NicolasCARPi:

PR #1113, opened 2015-08-01
Fix error when action called before initialized

### LINKED ISSUE #1112: Error when action called before the "data-api" initializes it
In the canonical Twitter Bootstrap plugins, any call initializes the plugin. This is done in this plugin too, but here it generates an error because the first argument is passed as the extra "event" parameter.

I've fixed it, I will submit a PR promptly.

mmacedo:
Fixes #1112.

As a bonus, since I was editing that part of the code, I removed the problematic "shift" and all its even uglier workarounds.


NicolasCARPi:

PR #1166, opened 2015-10-02
Changed the use of parentElement to parentNode.

farhan2106:
This needs to be changed because parentElement is not supported in FF <9.


caseyjhol:
The changes needed to be made to the source JS file. This was simple enough that I just did it myself. Thanks.

PR #1120, opened 2015-08-08
Add bidirectional CSS support

elad:
Allow both left-to-right and right-to-left layouts using a single CSS file.

Supersedes #572.

PR #1176, opened 2015-10-17
Added Right-to-Left support for languages like Arabic, Hebrew, etc

linostar:
Right-to-Left (RTL) languages need text aligned to the right (instead of the left), and carets/arrows/buttons to be displayed on the other side.

This commit fix the style so the select box appears properly with RTL pages (i.e. when dir="rtl" attribute is used in an HTML tag).


linostar:
This also covers some cases that pull request #1120 doesn't cover, like when the `dir` attribute is used in the `body` tag or any other tag, and not in the `html` tag.

NicolasCARPi:

Issue #1179, opened 2015-10-22
Customizing the text when multiple options are selected [implement doesn't works]


C0ZEN:
I've found this question : #797 which is now close.
I've got a similar question/request.
I tried to change the text display by creating some new option like that : 

    itemsSelected: null,
    countSelectedText: function (numSelected, numTotal) {
                return (numSelected == 1) ? '{0} item selected' : '' + itemsSelected + '';
    },
And of course call and implement it from js script like that : 

    $('.selectpicker').selectpicker({
        itemsSelected: '<?php echo ITEMS_SELECTED; ?>' // value change with language
    });
But it doesn't work. The value don't change. As far as I know the only way which is working is the official one : a simple string.

    // All KO
    return (numSelected == 1) ? '{0} item selected' : '' + itemsSelected + '';
    return (numSelected == 1) ? '{0} item selected' : itemsSelected;
So if someone can tell me how can I handle that and maybe add this option officially for countSelectText and maxOptionsText it will be wonderful.
Thanks.

caseyjhol:
What are you trying to do?

C0ZEN:
I just need to change the text if the language change (like you see from php constants). So I use selectpicker's options do to that but it seems that the return statement doesn't works in the list of options if I put an other type of value than a simple string... In other part of the .js function, we use a simple 'this.options.var' to display them but it is not working in countSelectedText because it's in the options part.  
So I just want to know if there is an other solution to do it directly from the options declaration of if I need to do it later in the code - and if it's later in the code, where ?!,- .

C0ZEN:
Ok I found the solution.

caseyjhol:
You need to also set `selectedTextFormat` to `'count'` or `'count > x'` to use countSelectedText.

PR #1183, opened 2015-10-23
Added new text if all items are selected

C0ZEN:
Just change the text if all items are selected.


vsn4ik:
LGTM

C0ZEN:
It works but if you want to change the text display with an option I don't if it's possible directly from here.
I asked about that #1179

C0ZEN:
The problem is solved in part. More about the next issue here #1053

PR #1188, opened 2015-10-30
Added options for all type of selected text

C0ZEN:
Added three options to change the text 'X item selected', 'X items selected' and 'all text selected'.
You can now change the value for each language for example when calling your select in js.

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -271,7 +271,12 @@
     noneSelectedText: 'Nothing selected',
     noneResultsText: 'No results matched {0}',
     countSelectedText: function (numSelected, numTotal) {
-      return (numSelected == 1) ? "{0} item selected" : "{0} items selected";
+      if(numSelected == numTotal){
+        return 'All items selected';
+      }
+      else{
+          return (numSelected == 1) ? '{0} item selected' : '{0} items selected';
+      }
     },
     maxOptionsText: function (numAll, numGroup) {
       return [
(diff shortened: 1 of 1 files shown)


caseyjhol:
Simply set a `countSelectedText` function instead:

```js
    countSelectedText: function (numSelected, numTotal) {
      if (numSelected == 1) {
        return "{0} item selected";
      } else if (numSelected === numTotal) {
        return "All items selected";
      } else {
        return "{0} items selected";
      }
    }

**Comment by hossein** [comment]
@caseyjhol Since the feature to show 'All items selected' was requested quite a few times, don't you think it's better to implement it (IMO as the default)? Without the need to roll your own `countSelectedText` function and simulate the `value > x` behavior additionally.

'All items selected' is way more descriptive than '1153 items selected'.

In the same way we have a `noneSelectedText` option we can have an `allSelectedText` option too.

**Comment by caseyjhol** [comment]
@hossein It probably would be a good idea. I'm looking to improve on the language options (and deprecate the existing), so it would it be easier to customize without needing dozens of data attributes.

**Comment by cameronmlewis** [comment]
Has something changed since this original post?

I'm trying to use what was mentioned above, but it doesn't seem be doing anything, here's what I'm using.

```
$('.my-select').selectpicker({ 
countSelectedText: function (numSelected, numTotal) {
      if (numSelected == 1) {
        return "{0} properties selected";
      } else if (numSelected === numTotal) {
        return "All properties selected";
      } else {
        return "{0} properties selected";
      }
    }
});
```

Is there any reason this shouldn't work?

hossein:
@cameronmlewis [As per the docs](https://developer.snapappointments.com/bootstrap-select/options/), `selectedTextFormat` should also be set to `count` or `count > #` for this to work.

cameronmlewis:
@hossein I've used `data-selected-text-format="count"` on the select item, but this hasn't actually done anything either, it still says x items selected?

PR #1201, opened 2015-11-08
Added Box Shadow CSS

rodrigoslayertech:


caseyjhol:
Extraneous CSS styling should be left up to each individual user. Unless there's something I'm missing?

PR #1206, opened 2015-11-18
Added option to let user to set menu width

juneidy:
The only file changed is js/bootstrap-select.js Everything else are changed due to I'm using linux environment which replaced \n\r to \r during build


NicolasCARPi:

PR #1237, opened 2015-12-15
Added Option to close Dropdown on Multiple Select

srd2010:
Needed an option to close the dropdown on multiple select boxes

Kept default behavior to stay open
Setting Option: "closeMultipleOnSelect" to true enables closing on selection


caseyjhol:
This seems like poor UX to me, but I'm willing to consider this option if somebody can make a case for it.

PR #1247, opened 2015-12-29
Update bootstrap-select.js

m0zart89:


caseyjhol:
Can you link to an example where the current method causes problems?

PR #1250, opened 2015-12-31
Restore select to its original position when destroying

jkrehm:
In my opinion, when destroying the bootstrap-select element, it should restore the select to its original position. In 1.7.7 (where I first noticed the issue) it was only an issue with the mobile version, but since `this.$element` is now always appended to `this.$newElement`, it's an issue in general.

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -1592,7 +1592,7 @@
     },
 
     destroy: function () {
-        this.$newElement.remove();
+        this.$newElement.before(this.$element).remove();
 
         if (this.$bsContainer) {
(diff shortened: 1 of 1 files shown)


jkrehm:
To be clear on the issue, currently when you fire the `destroy` function, it destroys not only the bootstrap-select stuff, but the original `<select>` element, too. I don't think that's the intention of `destroy`, so this fixes it.

caseyjhol:
Good catch. I'll test and merge this tomorrow.

jkrehm:
Awesome, thanks for merging.

machadolab:
This fixed my issue too! Thanks!

PR #1251, opened 2016-01-02
Add option to define own searchNormalizer

jansarmir:
The basic normalizeToBase is very simple and doesnt handle lot of EasternEurope diacritics.
With this option it would be easy to extend this basic normalizer.

Like [Remove diacritics](https://github.com/jansarmir/Remove-Diacritics)


Stichoza:
This feature would be really useful. I want to add Georgian letters mapping because some users type with latin letters ("xinkali" instead of ხინკალი).

NicolasCARPi:

PR #1283, opened 2016-02-04
wai-aria and accesiliby fixes

neo87cs:
The select should be roled as "listbox" and the options as "option", this fix make the select respect the WAI-ARIA 1.0 standard. Added an hook for "enter key" pressed, so the component can be used with keyboard too.

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -451,7 +451,7 @@
           header +
           searchbox +
           actionsbox +
-          '<ul class="dropdown-menu inner" role="menu">' +
+          '<ul class="dropdown-menu inner" role="listbox">' +
           '</ul>' +
           donebutton +
           '</div>' +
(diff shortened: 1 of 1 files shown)


caseyjhol:
This looks good, except for the keypress event. I'm about confused by it. There's already a keydown function that allows the plugin to be used with a keyboard. Unless there's some functionality it's providing that I'm missing?

neo87cs:
The plugin is based on click event, but when the select is expansed the
enter key is Not catched. What i did is catch keypress event and throw a
simple click (i am a java developer, i am not expert With js). For
accessibility i simple fixed the tag generation.
Il 04/feb/2016 08:54 PM, "Casey Holzer" <notifications@github.com> ha
scritto:

caseyjhol:
You can already select options using the Enter key with the current version.

neo87cs:
Aw, i had 1.9.4 (stable one), and it didn't work With enter key. I knew it
cause i developed accessibility. I hadn't the time to test it in v2 because
we are in production.
Il 04/feb/2016 09:33 PM, "Casey Holzer" <notifications@github.com> ha
scritto:

caseyjhol:
It should be working in v1.9.4. What browser/OS are you using that it's not working for you?

neo87cs:
Chrome on Ubuntu 15.10
Il 04/feb/2016 09:58 PM, "Casey Holzer" <notifications@github.com> ha
scritto:

caseyjhol:
Hrmmm...it's working for me on Chrome 48 on Ubuntu 15.04. On the examples page (http://silviomoreto.github.io/bootstrap-select/examples/), opening the first select and then using the arrow keys to highlight an option and pressing "Enter" doesn't work for you?

PR #1298, opened 2016-02-16
reset scroll position whith live-search

kayun:
resets scroll position only singleselect whith live-search when you click on an option

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -1304,7 +1304,9 @@
         }
 
         that.$lis.filter('.active').removeClass('active');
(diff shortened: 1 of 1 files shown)


caseyjhol:
Closing this until somebody can explain what the problem is or provide an example.

andie-chernysh:
@caseyjhol, not sure if that's the same case, but when I select something in selectpicker dropdown and try to reset my selection with `element.val('').selectpicker('refresh')` - it resets only selected value, but not the scroll position in the dropdown. `.selectpicker('render')` didn't work either. I even don't know what `.selectpicker('render')` does. It looks like it does nothing.... Not re-rendering the picker layout - that's for sure.

PR #1326, opened 2016-03-10
Small tweak to avoid empty initial element creation

ghost:
Hi there! We are using your plugin and we found an scenario where once you select any option from the dropdown something happens and later on when you go back to the page with the dropdown, when you want to select another option, an empty element is being created and inserted at the very beginning of the list of options. 

As you will see in the PR (please only pay attention to the if block in the generateLI method, the rest should be just indentation), the tweak is minor and we are pretty sure that you will find a more neat and elegant solution to this bug.

Please let us know when you release the fix since we need due to already have a running version in production.

Thanks in advance.

Agustín Cassani

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -111,15 +111,15 @@
 
   if (!Object.keys) {
     Object.keys = function (
-      o, // object
-      k, // key
-      r  // result array
-      ){
+        o, // object
+        k, // key
+        r  // result array
+    ){
       // initialize object and result
       r=[];
       // iterate over object keys
       for (k in o)
-          // fill result array with non-prototypical keys
+        // fill result array with non-prototypical keys
         r.hasOwnProperty.call(o, k) && r.push(k);
       // return result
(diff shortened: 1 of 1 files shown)


caseyjhol:
Could you link me to a demonstration of this problem?

surfercoder:
Desar Casey,

Thanks a lot for your quick response. Since you need to be logged in in the
beta version we have in production, I have recorded how to reproduce the
scenario. I hope it helps.

But a brisk explanation would be the following: We are going to an specific
page, then to the advance settings, we apply some filters where is the
bootstrap-select applied, the first time works perfectly, then we reset the
filters and go back to the previous page and when we want to apply another
filter in the same dropdown, the empty element was inserted at the very
beginning of the dropdown.

Let me know if you need anything else from our side.

Best regards.

Agustín

caseyjhol:
The issue you're describing may have been fixed in a more recent version. In any case, this proposed change would cause the first option to never be rendered.

PR #1335, opened 2016-03-24
setStyle: remember previous style

malloctrim:
Repeatedly calling `setStyle(newStyle)` does not work, as `removeClass`only removes the original `this.options.style` value.

`this.options.style` should be set to `buttonClass`.


NicolasCARPi:

PR #1356, opened 2016-04-08
Use undefined object and jQuery trigger method

vsn4ik:
1. Use native focus method if possible
2. http://jsperf.com/is-undefined
3. Use `$.fn.trigger` for unification. Deprecate event shorthand methods (jQuery v4, https://github.com/jquery/jquery/issues/3214)

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -388,11 +388,11 @@
 
       if (this.options.dropdownAlignRight === true) this.$menu.addClass('dropdown-menu-right');
 
-      if (typeof id !== 'undefined') {
+      if (id !== undefined) {
         this.$button.attr('data-id', id);
-        $('label[for="' + id + '"]').click(function (e) {
+        $('label[for="' + id + '"]').on('click', function (e) {
           e.preventDefault();
-          that.$button.focus();
+          that.$button.trigger('focus');
         });
       }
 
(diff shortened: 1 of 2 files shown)


js/bootstrap-select.js
```diff
@@ -558,16 +557,16 @@
...
+            icon = $this.data('icon') !== undefined ? '<span class="' + that.options.iconBase + ' ' + $this.data('icon') + '"></span> ' : '',
             isOptgroup = this.parentNode.tagName === 'OPTGROUP',
             isDisabled = this.disabled || (isOptgroup && this.parentNode.disabled);
 
         if (icon !== '' && isDisabled) {
           icon = '<span>' + icon + '</span>';
         }
 
-        if (that.options.hideDisabled && (isDisabled && !isOptgroup || this.parentNode.disabled && isOptgroup)) {
+        if (that.options.hideDisabled && isDisabled) {
```
caseyjhol: This line needs to be `that.options.hideDisabled && (isDisabled && !isOptgroup || this.parentNode.disabled && isOptgroup)`. With just `that.options.hideDisabled && isDisabled`, if an optgroup has a disabled option in it, the optgroup's header is hidden. I guess we could make a variable called something like `isOptgroupDisabled` and set it equal to `this.parentNode.disabled && isOptgroup`, and then set the if statement to `if (that.options.hideDisabled && (isDisabled && !isOptgroup || isOptgroupDisabled))`.

vsn4ik:
@caseyjhol better looks now?

js/bootstrap-select.js
```diff
@@ -598,19 +598,14 @@
             _li.push(generateLI(label, null, 'dropdown-header' + optGroupClass, optID));
           }
 
-          if (that.options.hideDisabled && isDisabled) {
-            liIndex--;
-            return;
-          }
-
```
caseyjhol: This needs to remain, otherwise disabled options within an optgroup stay visible.

vsn4ik: Done.

PR #1379, opened 2016-05-03
Allow HTML in title

chandon:
Default title option doesn't allow HTML rendering. All html tags are converted to text.

This fix the issue


golear91:
appreciate this, I needed this myself.

caseyjhol:
Seems like HTML is already allowed. Can you provide an example?

Issue #1159, opened 2015-09-16
Collapsible Opt Groups


vatsalpande:
Hi,

When using bootstrap select I think it will be great if we have an option to expand collapse optgroups. This way if there are multiple records it will be easy for the user to just go to the desired category rather than scrolling all the way down or up.

Please suggest if its in scope

smisch:
+1

Meanwhile you could easily implement this feature:

````
$(function (){
	// for some strange reason selectpicker prevents the click-event. so just use mouseup
	// when clicking on an optgroup "label", toggle it's "children"
	$(document).on("mouseup", ".bootstrap-select .dropdown-header", function (){
		var $optgroup = $(this),
			$ul = $optgroup.closest("ul"),
			optgroup = $optgroup.data("optgroup"),
			
			// options that belong to this optgroup
			$options = $ul.find("[data-optgroup="+optgroup+"]").not($optgroup);
		
		// show/hide options
		$options.toggle();
		
		$optgroup.toggleClass("closed");
	});
	
	// initially close all optgroups that have the class "closed"
	$(document).on("loaded.bs.select", function (){
		$(this).find(".dropdown-header.closed").each(function (){
			var $optgroup = $(this),
				$ul = $optgroup.closest("ul"),
				optgroup = $optgroup.data("optgroup"),
				
				// options that belong to this optgroup
				$options = $ul.find("[data-optgroup="+optgroup+"]").not($optgroup);
			
			// show/hide options
			$options.toggle();
		});
	});
});
````

````
.bootstrap-select .dropdown-header {
	cursor: pointer;
}
````

vatsalpande:
Hi,
Thanks a lot for your reply :)

I was playing around with bootstrap-select for a while as we are using it heavily in our project.
I have made few smaller enhancements to it -
a) Gives you collapsible configurable dropdowns
b) Gives you the option of having custom text in place of item selected. For ex - 4 countries selected, 5 toys selected. And also gives the option of configuring text when all items are selected.

Not sure if its of any other persons use so didnt share as it may be specific custom project implementation

Cheers

vatsalpande:
An example of what I did can be found at 

http://stackoverflow.com/questions/32571525/accordion-inside-bootstrap-select/33731926#33731926


https://jsfiddle.net/zqvvuts6/5/

wenfei-qi:
Great stuff ! Thanks ! @vatsalpande

vatsalpande:
[...]

PR #1402, opened 2016-05-28
Create a ul for each optgroup

### LINKED ISSUE #1401: Suggested change to menu structure for optgroups
I'm wanting to style each optgroup a certain way but the current html structure doesn't allow me to do that. My suggestion is to structure it like this instead, with each optgroup a separate ul:

`<div class="dropdown-menu">
    <div class="dropdown-header">...</div>
    <ul class="dropdown-menu inner">
        <li>...</li>
    </ul>
    <div class="divider"></div>
    <div class="dropdown-header">...</div>
    <ul class="dropdown-menu inner">
        <li>...</li>
    </ul>
</div>`

No changes would be needed CSS wise, the appearance remains identical to how it looks currently.

andrews05:
This change creates a `ul` wrapper around all `li`s associated to an optgroup, allowing the optgroups to be targeted/styled using the `.optgroup` class.
Fixes #1401, should also make it easier to do things such as select/deselect all (#737) or collapsing (#1159).

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -288,6 +288,12 @@
       this.options.title = this.$element.attr('title');
     }
 
+    // Format window padding
+    var winPad = this.options.windowPadding;
+    if (typeof winPad === 'number') {
+      this.options.windowPadding = [winPad, winPad, winPad, winPad];
+    }
+
     //Expose public methods
     this.val = Selectpicker.prototype.val;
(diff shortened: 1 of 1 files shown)


andrews05:
Any thoughts on this? I've discovered that using this actually makes a noticeable improvement to rendering performance for large lists, despite being no different in appearance.
It's a real simple tweak that doesn't break anything else.

PR #1418, opened 2016-06-16
Fix wai-aria and accessibility violations

sanmuny:
When using IBM Rational Policy Tester to test the accessiblity, 4 violations
are found:
1. Each form control must have associated text.
2. The WAI-ARIA property aria-expanded is not supported global property on
   element BUTTON that does not have a WAI-ARIA role.
3. The descendent element with WAI-ARIA role does not have tabindex attribute
   set to 0 or -1 to be accessible by keyboard.
4. A WAI-ARIA widget must have an accessible name.

This patch fixed these violations.
Note: pull request #1283 is also included.

Signed-off-by: Wang Sen wangsen@linux.vnet.ibm.com


okidokiPVD:
+1 to this. Features better compliance.

Issue #1425, opened 2016-06-22
Remove additional markup per config


anthonygreco:
When setting ```showTick: true```, the ```<span class="glyphicon glyphicon-ok checkmark"></span>``` markup is still rendered. Another ternary conditional in the ```generateA``` method resolves this. PR coming.

caseyjhol:
Released in [v1.13.3](https://github.com/snapappointments/bootstrap-select/releases/tag/v1.13.3)!

PR #1426, opened 2016-06-22
conditionally render the tickIcon markup based on options.showTick

anthonygreco:
Re: Issue https://github.com/silviomoreto/bootstrap-select/issues/1425


caseyjhol:
This doesn't take into account the possibility of a multiple select.

PR #1429, opened 2016-06-24
Updated style sheet for Bootstrap Input Sizing issue

shailesh-daund:
Currently if you add any Bootstrap Input Sizing class with select picker it will not adjust the size accordingly.

I have updated the style sheet to resolve this issue.

**Currently it looks like**

select_picker_with_input-lg_current

**After this updates it will looks like** 

select_picker_with_input-lg_after_update


NicolasCARPi:

PR #1439, opened 2016-07-15
minimumInputLength option

nemiro:
added minimumInputLength feature.
message can be specified at options as string, or as a function.

allow to specify minimum input length to show results


NicolasCARPi:

PR #1472, opened 2016-08-17
Add the ability to select/deselect all options in an optgroup by clicking on header

Flaque:
- Ref #737 
- Uses a data-attribute: `data-select-all-headers="true"`
- Slightly modifies the CSS to make headers seem selectable.
- Added to Docs in Options and Examples.
- Added example to test.html

---
### Fills completely on partially checked option groups

If the use has only selected 3/5 of the total possible options in an `optgroup` it will select all 5/5. I figure it is better to add things to the end user than to take away their work. People tend to be more attached to things they've done than things that they feel were done by a computer. 
### Selects the first few if `maxOptions` is set

I incorporated the idea from @weatherbell to select the first `maxOptions`. I like this idea a lot because if the user is confused, the first thing the user will try to do is select the next few which will then force the warning: `Group limit reached ({n} item max)`. So any confusion is self-fixing. 
### It's a data attribute

In order for the programmer to add the "select header" functionality, they must add the `data-select-all-headers="true"` attribute to their `<select>` like so:

```
<select id="selectHeader" class="selectpicker form-control" multiple data-select-all-headers="true" title="Selecting a header will select all it can.">
<!-- Options with groups -->
</select>
```
### Documentation

Requiring this to be opt-in means the user will have to read the documentation in order to add it to their project. @Khrysller was right that we should add documentation as a solution to the maxOptions problem.


caseyjhol:
The maxOptions issue was always more of a UX problem for me...meaning that I could see the actual users being confused by the seemingly inconsistent functionality. However, if the developer is aware of the issue and doesn't mind it (for example, in a situation where they don't even utilize maxOptions at all), then I think it is a good feature to have. I'll review your code and get back to you with any updates.

Flaque:
@caseyjhol Heads up, I've realized that selecting a header doesn't properly fire the jQuery change event because it's never hitting a `.triggerNative('change')` [like it does here](https://github.com/silviomoreto/bootstrap-select/blob/master/js/bootstrap-select.js#L1315-L1322). If no one objects, I'm going to just call the `.click()` event for each option instead of manually doing it. Working on that now.

KrunchMuffin:
Will this ever be integrated?

mikenorm412:
Any plans on integrating these changes?  Is there a workaround available to gain this functionality or at least trap the click event on the optgroup?

Flaque:
Pretty sure this will not be integrated, considering it's been in an open PR for a year. Best solution is probably to fork it.

timammons:
This change would be really helpful!  Can this be integrated please?

shofer4eto:
> The maxOptions issue was always more of a UX problem for me...meaning that I could see the actual users being confused by the seemingly inconsistent functionality. However, if the developer is aware of the issue and doesn't mind it (for example, in a situation where they don't even utilize maxOptions at all), then I think it is a good feature to have. I'll review your code and get back to you with any updates.

Any idea if this will make it into the core? Would be a great addition to an already awesome package! Thanks!

caseyjhol:
@shofer4eto See #737. Please 👍 to vote and subscribe 🔉 for updates.

PR #1478, opened 2016-08-21
take searchHeight into account when calculating inner menu's max-height

suntribe:
In situations when size is set and liveSearch is also set to true, $menuInner max-height didn't take into account it's height and the max-height was wrong..


caseyjhol:
searchHeight isn't included as a part of menuHeight, which is why it needs to be added when setting $menu max-height. Subtracting searchHeight from $menuInner's max-height makes it too small. If you could show me an example I'll consider reopening/merging this.

Issue #1502, opened 2016-09-09
Add option to support window padding


birjolaxew:
In cases where the window has areas that the dropdown menu should not cover - for instance a fixed header - it would be useful to have support for window "padding" to offset the `selectOffset*` calculations.

[JSFiddle of problem.](https://jsfiddle.net/d4uposgp/)

Image of problem:
<p align="center">
<img src="https://cloud.githubusercontent.com/assets/4542461/18387599/bda6f9e6-769d-11e6-8b07-2ac30525df04.png" />
</p>

birjolaxew:
Implemented it in a fork. Discussion takes place in #1503.

PR #1503, opened 2016-09-09
Added support for window padding

birjj:
As described in #1502.
Added an option called `windowPadding`, which supports the formats `x`, ~~`"x y z k"`~~ and `[x, y, z, k]`, where `x`, `y`, `z` and `k` are integers. The string and array formats use the normal CSS order (top-right-bottom-left), while the number format applies to all sides.

[Fixed JSFiddle.](https://jsfiddle.net/rvaq2a11/)

Image comparisons:

<p align="center">
  
  
</p>

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -269,6 +269,19 @@
       this.options.title = this.$element.attr('title');
     }
 
+    // Format window padding
+    var winPad = this.options.windowPadding;
+    switch (typeof winPad) {
+      case "number":
+        this.options.windowPadding = [winPad, winPad, winPad, winPad];
+        break;
+      case "string":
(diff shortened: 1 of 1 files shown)


birjolaxew:
Updated to remove `dist/*` from commit so it can be merged without conflicts

caseyjhol (changes requested):
Seems like a useful feature. I'm not sure we need to support the string option, though. Change it to support just a number or array of numbers using a simple `if (typeof winPad === 'number')` and I should be able to get this merged into the next version.

birjolaxew:
Done.

caseyjhol (approved):

vsn4ik:
It should be added in documentation.

caseyjhol:
Yes, I'll get it added before the official release.

PR #1514, opened 2016-09-15
Update bootstrap-select.js

kdcolvin:
the object "that" was not assigned sizeInfo but each sub element of sizeInfo directly.  Thus the correct code would be that.liHeight versus that.sizeInfo.liHeight.

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -1212,7 +1212,7 @@
 
           // scroll to selected option
           var offset = that.$lis.eq(selectedIndex)[0].offsetTop - that.$menuInner[0].offsetTop;
-          offset = offset - that.$menuInner[0].offsetHeight/2 + that.sizeInfo.liHeight/2;
+          offset = offset - that.$menuInner[0].offsetHeight/2 + that.liHeight/2;
           that.$menuInner[0].scrollTop = offset;
         }
       });


caseyjhol:
https://github.com/silviomoreto/bootstrap-select/blob/master/js/bootstrap-select.js#L869

?

kdcolvin:
Line 1215:   offset = offset - that.$menuInner[0].offsetHeight/2 + **that**.sizeInfo.liHeight/2;  //sizeInfo is undefined.  "that", not "this" is referenced.
My ref lines 888-928 where line 894: liHeight = this.sizeInfo['liHeight'], 
This causes an exception when adding the undefined value for that.sizeInfo.liHeight/2 to that.$menuInner[0].offsetHeight/2.

caseyjhol:
`that` is equal to `this`.

kdcolvin:
Hopefully a better explanation:

With the following selectpicker call on document.ready or on pageLoad() method the below issue occurs after the control loads and is displayed on page.  

  $('select').selectpicker({
                style: 'btn-primary',
                width: 'auto',
                liveSearch: 'true',
                liveSearchPlaceHolder: 'Type to Search'
            });

When executing click event on button, line 1215 fires that.sizeInfo.liHeight exception "0x800a138f - JavaScript runtime error: Unable to get property 'liHeight' of undefined or null reference".   sizeInfo is actually undefined at this point.   

If you continue execution after the above exception, the control functions as expected unless page is reloaded or no selection is made.  In either case, the exception fires again.

 A direct reference to that.liHeight on line 1215 returns no undefined exception.

Another issue:  The reason for placing selectpicker call in pageLoad was an attempt to resolve an issue with the style: 'btn-primary' property not being applied on page load and btn having class of btn-default instead of btn-primary.

I hope this helps explain the issue further.

bootstrap-select_exception

caseyjhol:
What version of Bootstrap are you using? I see your point (even if I can't recreate it), but using that.liHeight is not the solution. That is a reference to the liHeight function. The function divided by 2 does not equal the height. I'm also not sure how that is being called *before* the `setSize` function, as `setSize` is called on click, and the line you're referencing isn't called until the menu is shown.

Issue #1275, opened 2016-01-29
Live search slow in IE


Robelind:
Hi.
I'm populating the select with about 3000 entries and when using live search in IE11, the experience is quite bad. When writing e.g. 3 characters it takes literally seconds for the select to do the search.
Doing the same thing in Firefox shows acceptable performance.

Bootstrap-select 1.9.4, Windows 7, jQuery 2.14, Bootstrap 3.3.5

MOKuper:
I also am experiencing the same difficulty with a little over 4000 records. Chrome is really slow at live search, taking 6 seconds or more when fully typing a name. In firefox it is still slow, but remarkably better. 

Using Bootstrap-select 1.10.0, mac el captan, jquery 1.11.3, bootstrap 3.3.2

git-volver:
In fact, all browser I tested is so slow when the records are over 2 thousands... Big problem for my system...

Using Bootstrap-select 1.10.0, jquery 1.12.2, bootstrap 3.3.6

omggga:
Check function with delegating on "input propertychange" event.
There are lots of work with DOM: filtering (2 times), escape function, replacing and html inserting and other. If you have more than 1000 entries and some optgroups, it works very slow in all browsers.

The only way i see now is to set throttling and increasing live-search keyword length from zero to 3 or more. Thats not removing the problem, but resolving some user UI lag questions.

ataft:
Are there any updates on the performance issues?

I opened issue https://github.com/silviomoreto/bootstrap-select/issues/1229, suggesting virtualization when scrolling on large lists ([ag-grid](https://www.ag-grid.com/) does this very well). Perhaps ambitious, but having the ability to load from JSON (issue https://github.com/silviomoreto/bootstrap-select/issues/1151), live search against the JSON, and have virtual scrolling would solve all the performance issues and allow for very large lists. Perhaps some of that is covered by the Ajax Bootstrap Select clone, referenced in issue https://github.com/silviomoreto/bootstrap-select/issues/953, but I don't need ajax search. It would be great to not even have to have the SELECT in the DOM.

Again, this might be too ambitious, but would enable people to have as many bootstrap selects as they want with as many options as they want without any issues.

xssunircomercial:
Help me
https://github.com/silviomoreto/bootstrap-select/issues/1650

caseyjhol:
[...]

PR #1516, opened 2016-09-18
Improve Livesearch Performance

andrews05:
When working with large lists, livesearch can start to show a little lag when typing. I've reworked the input listener to improve the performance. The gains are primarily from reducing the number of queries that are made. In my test case, this code runs at least 10x faster than before.
This also fixes a couple of edge cases where a divider wouldn't show/hide correctly.

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -1430,58 +1430,51 @@
       });
 
       this.$searchbox.on('input propertychange', function () {
+        that.$lis.not('.is-hidden').removeClass('hidden');
+        that.$lis.filter('.active').removeClass('active');
+        $no_results.remove();
+
         if (that.$searchbox.val()) {
-          var $searchBase = that.$lis.not('.is-hidden').removeClass('hidden').children('a');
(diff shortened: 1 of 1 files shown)


andrews05:
Minified files removed as per updated guidelines.
Keen to hear any comments on this as it makes a real difference to how responsive it feels in some of the menus I'm using.

poziomek:
1. List with 3600 entries, both: values and displayed text are strings. Changes from this commit help us to reduce search latency (during typing). Great work.
2. I've took source codes directly from commit (bootstrap-selec.js file from commit window) and it seems that works. 
However when I took minified version from dist folder (don't remember exactly)  then error raises (script doesn't compile) in browser.

caseyjhol:
Will try to get this implemented ASAP.

poziomek:
Thank you!

caseyjhol:
This also addresses #1275.

andrews05:
Thanks for merging this! Glad to hear I'm not the only one working with stupidly large lists.

Issue #1193, opened 2015-11-03
'title.replace' issue


aydn:
![Capture 1](http://i.hizliresim.com/p9MrAo.png)
-
![Capture 2](http://i.hizliresim.com/kJ7VQJ.png)

IDK why but I think $.attr returns Number instead of a String type

miqmago:
This happens when a numeric value is assigned as the selectpicker value:

```js
$('.mySelectPicker').selectpicker();
$('.mySelectPicker').val(1);
$('.mySelectPicker').selectpicker('refresh');
```

PR #1519, opened 2016-09-20
'title.replace' issue when data-content is a number

miqmago:
https://github.com/silviomoreto/bootstrap-select/issues/1193

``` js
$('.mySelectPicker').selectpicker();
$('.mySelectPicker').val(1);
$('.mySelectPicker').selectpicker('refresh');
```

Produces the following error:

``` js
angular.js:13920 TypeError: title.replace is not a function
    at Selectpicker.render (bootstrap-select.js:793)
    at Selectpicker.refresh (bootstrap-select.js:1772)
    at HTMLSelectElement.<anonymous> (bootstrap-select.js:1843)
    at Function.each (jquery.js:384)
    at jQuery.each (jquery.js:136)
    at jQuery.Plugin [as selectpicker] (bootstrap-select.js:1823)
    at bootstrapSelectpicker.js:18
    at angular.js:19612
    at completeOutstandingRequest (angular.js:5964)
    at angular.js:6243
```

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -19,8 +19,7 @@ Guidelines for bug reports:
 4. **Create an isolated and reproducible test case.** Create a [reduced test
    case](http://css-tricks.com/6263-reduced-test-cases/).
 
-5. **Include a live example.** Make use of jsFiddle or jsBin to share your
-   isolated test cases.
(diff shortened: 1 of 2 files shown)


caseyjhol:
I can't recreate this issue you're referring to. Can you link me to an example?

miqmago:
https://jsbin.com/vilebefuqe/edit?html,js,console,output

The error is produced when using an integer value with `data-content`.

caseyjhol:
I see. This is actually an issue when data-content is a number, and doesn't have anything to do with the value, which is why I wasn't able to recreate it. Line 760 is all that needs to be changed.

`return $this.data('content');` to `return $this.data('content').toString();`

Change that in js/bootstrap-select.js (and remove all other changes) and I'll get this merged.

PR #1524, opened 2016-09-22
Update select button's `aria-disabled` attribute when the button is disabled.

geramirez:
- This change will make VoiceOver inform the user the button is disabled


caseyjhol:
You'll also need to add `.attr('aria-disabled', false);` on line 1185.

PR #1540, opened 2016-10-20
Move "dropdown-toggle" into styleBase-option

rejas:
I was trying to overwrite all styles for the select-button. Unfortunatly "dropdown-toggle" was added to the button anyway, so I looked and found the (undocumented) styleBase option.
So I added a documentation for it and moved the dropdown-toggle class into it. 

This MIGHT break for people that already use the styleBase-option and depend on the dropdown-toggle-styling so I am more then eager to hear your opinion on this.

### DIFF
--- a/docs/docs/options.md
+++ b/docs/docs/options.md
@@ -241,9 +241,18 @@ Options can be passed via data attributes or JavaScript. For data attributes, ap
   <tr>
     <td>styleBase</td>
     <td>string</td>
-    <td><code>btn dropdown-toggle</code></td>
+    <td><code>btn</code></td>
     <td>
       <p>Base style for all buttons.</p>
+    </td>
+  </tr>
+  <tr>
+  <tr>
+    <td>styleDropdown</td>
+    <td>string</td>
+    <td><code>dropdown-toggle</code></td>
+    <td>
+      <p>Dropdown style for all buttons.</p>
     </td>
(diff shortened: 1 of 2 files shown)


caseyjhol:
The `dropdown-toggle` class is used throughout the bootstrap-select CSS and is necessary for proper styling, meaning it is a necessary class and should not be optional. Also, as you mention, it would cause a breaking change for anybody already using the styleBase option. I do agree there needs to be documentation for styleBase, though. Making the dropdown-toggle class optional will probably have to be addressed in the next major release.

rejas:
I moved "dropdown-toggle" into a new option so that existing users wont experience a breaking change whiel stil lallowing people (like me) to change it. Hope you are okay with it.

rejas:
And thx for the awesome plugin by the way :-)

caseyjhol:
Can I ask - why do you want to change the class name? I'm just not sure I see this being a popular enough request to justify including it.

rejas:
In one of my projects we use components that have their own ui (and dont use any of the bootstrap styles or themes). So the boostrap-select buttons look different than the other buttons and overwriting the bootstrap styles turned out to be more difficult and hacky then just replacing the default "btn"-class. So for unifying the styles of non-bootstrap and bootstrap-select buttons this change would greatly help me.

PR #1545, opened 2016-10-28
mistaken range split into separate chars, polish diacritics

pixsrv:
fix: mistaken range split into separate chars
add: polish diacritics


caseyjhol:
The normalizeToBase function has been revamped for the next release, and this issue should be resolved.

PR #1551, opened 2016-11-09
Advanced normalizeToBase function which handles more UTF characters.

lukaszdechovan:
I implemented bootstrap-select in application localized to slovak language. In slovak it is possible to have a word "Halušky" which should be interpreted in accent insensitivity way as "Halusky". Your normalizeToBase function was handling only a few characters, so I created a new one which handles more characters.

Then I realized that to pseudo attributes for search was handling "li" instead of "a" elements and therefore cannot find data-tokens and data-normalized-text attributes. Therefore I changed these pseudo attribute functions to work with children "a" attributes.


caseyjhol:
A more advanced normalizeToBase function will be used in the next release. In the future, please follow contributing guidelines.

caseyjhol:
Released in [v1.13.4](https://github.com/snapappointments/bootstrap-select/releases/tag/v1.13.4)!

PR #1562, opened 2016-11-22
Add a `liveSearchNumber` option

### LINKED ISSUE #1492: Change live search's input type
# Need/use case

Using a Bootstrap select for a combo with years only, live search enabled. On mobile, the full keyboard is displayed.

We wish to specify the input is numeric and therefore the mobile device to display only the numeric/phonepas keyboard.

# Possible leads for solution

## input's `type` attribute

Allowing to change the input HTML5 type (such as [tel](http://mobileinputtypes.com/#tel)). [A `liveSearchType` option](https://silviomoreto.github.io/bootstrap-select/options/) could do the trick.

## input's `inputmode` attribute

Seems more complicated and [not sufficiently supported yet](http://caniuse.com/#feat=input-inputmode).

----

### Side notes:

1. As an IT guy, I hate it when a user gives me the solution they imagine instead of what they need, as what they ask is often ill-fitted to what they really want/need. My use case and need are specified, the solution is up to you!
2. I'm willing to do this myself if I chance to find the time, which won't be in the coming weeks but probably before the end of September.

### Related:

- [StackOverflow question](https://silviomoreto.github.io/bootstrap-select/options/)

cyChop:
Switches the input type from `text` to `tel`. Mobile devices will display only a numeric keypad instead of a full keyboard.
Fixes #1492.

Chose against leaving `liveSearchType` fully free in order to ensure correct use.
Possible enhancement: allow `email` input type. Not sure about the added value (I think this is less common as year or numeric reference selection).

### DIFF
--- a/docs/docs/options.md
+++ b/docs/docs/options.md
@@ -126,6 +126,14 @@ Options can be passed via data attributes or JavaScript. For data attributes, ap
     </td>
   </tr>
   <tr>
+    <td>liveSearchNumber</td>
+    <td>boolean</td>
+    <td><code>false</code></td>
+    <td>
+      <p>When set to <code>true</code>, the input type of the search field will be changed so that a numeric keypad is displayed on mobile devices instead of a full keyboard.</p>
+    </td>
+  </tr>
+  <tr>
     <td>maxOptions</td>
     <td>integer | false</td>
     <td><code>false</code></td>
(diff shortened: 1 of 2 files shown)


caseyjhol:
I'm thinking a better alternative might just be to support setting the `liveSearch` option to a string. If it's a string, then the input type is set to that.

```html
<select data-live-search="tel">
...
</select>
```

caseyjhol:
Note: this addresses #1492.

NicolasCARPi:

Issue #1587, opened 2016-12-11
Accent insensitive not working with liveSearchNormalize property


naisonsouza:
**See this fiddle example**: [https://jsfiddle.net/53wy5czj/2/](https://jsfiddle.net/53wy5czj/2/)
Search "vito" or "vitó" no results!

caseyjhol:
This is fixed in master via commit ba65054cc966da18634f14a7215840e3e24edc31.

See https://jsfiddle.net/53wy5czj/3/.

PR #1597, opened 2016-12-26
Fix selector query normalizedText when liveSearchNormalize=true

naisonsouza:
Issue #1587


caseyjhol:
This is fixed in master via commit ba65054cc966da18634f14a7215840e3e24edc31.

See https://jsfiddle.net/53wy5czj/3/.

Issue #1615, opened 2017-01-26
Performance problems with live search (update)


RvaDev2017:
Hello,

we have seen there several issues opened in regard to this issue. It seems that it was solved in a past  revision but the problem reappeared with commits of later revisions and it´s still happening in the latest versión available. 

* Environment details : Explorer and Chrome (although Chrome performs slightly faster than IE), Windows 7, Web under IIS 6.2, Bootstrap v3.3.4, bootstrap-select v1.12.1, jQuery v2.1.1

* Steps to reproduce : 
1) Create a selectpicker containing 1000 elements. Each element will have the same Text and Value and will be strings of 50 chars in length.
2) Options on selectpicker creation :  @data_live_search = "true", @data_live_search_normalize = "true", @data_select_on_tab = "true", @data_dropup_auto = "false" })
3) Access the select picker via chrome o IE and use live search to look for a certain entry.
4) Check that performance both on writting in the livesearch area and the search itself are both very slow.

In regard to this issue and scenario, we suggest to create a new creation option which allowed to perform live searches only when a minimun number of characters have been written in the search area. This will allow the developer to decide when to perform the live search allowing him to increase performance avoiding unneccesary searches when only a few characters have been typed in the live search area.

Thanks in advice.
Best regards.

Issue #1606, opened 2017-01-09
LiveSearch performance


andrews05:
The fix implemented for the key word searching (ba65054cc966da18634f14a7215840e3e24edc31) unfortunately made the performance much worse again. The culprit is the .parent() function - somehow over half the performance improvement from #1516  was gained by the removal of .parent().
The best solution I can find is to move the `.find('a')` into the pseudo expressions, which allows to remove .parent() while still keeping the key words working. I was thinking there could be remove for some extra improvement as well though. Here's what I have in mind:

1. Move .find('a') into the pseudo expressions as mentioned above.
2. Use the token to store normalized text, eliminating the need for the extra pseudo expressions.
3. Use normal functions instead of pseudo expressions and optionally accept a function for the liveSearchStyle. This is marginally faster and would allow users to implement their own search function which could be used to do something like 'word starts with', or to provide optimal performance by ignoring tokens.

Any thoughts on this? I'm happy to make a PR for this if you think it would be a good idea.

(P.S. Much apologies for breaking the key word searching)

caseyjhol:
I'm working on an updated version of bootstrap-select that uses virtualization (rendering only the options that are actually visible). This makes live search *instant* (using a select with 4000 options in my testing) and works pretty awesome. The slowest part about the search currently is showing/hiding thousands of elements at once. I'm working on making this new version available in a separate branch so others can test it as well.

andrews05:
Awesome, I look forward to testing that out :)

caseyjhol:
I just added a virtual scrolling feature (which results in instant searching) in the [virtual-scrolling](https://github.com/silviomoreto/bootstrap-select/tree/virtual-scrolling) branch. It's still a WIP, and there are a few features missing, some I'm aware of (the startsWith feature for example), and some I'm sure I'm not. Try it out - I'm looking for feedback. Please post all findings in #1229.

PR #1631, opened 2017-02-08
Livesearch performance

andrews05:
This change once again makes significant improvements to the performance of livesearch by eliminating the call to .parent(), this time without breaking keyword searching. See #1606 for more info.
This should also address #1615.

I also made it always scroll to the top of the list when you type. This was previously done by using .focus() on the top item but my last PR removed this for performance reasons (I didn't realise what purpose it was serving). Using .scrollTop() achieves the same result without the performance cost.

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -1458,6 +1458,7 @@
             if ($foundDiv) $foundDiv.addClass('hidden');
 
             $searchBase.not('.hidden').first().addClass('active');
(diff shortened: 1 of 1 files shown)


caseyjhol:
Looks good!

PhilPJL:
Sounds great.  Could really do with this fix since we have > 1000 items in the picker.  Any idea when it might be released?

gmsa:
Would love a release with this fix as well

PR #1635, opened 2017-02-12
Fix positioning issue on panned/zoomed viewport on chrome with container: body

jesenko:
jQuery `offset()` seems to returns wrong absolute position in chrome when viewport is panned/zoomed, resulting in wrongly positioned bootstrap select with `container` set. Alternative method using standardized `getBoundingClientRect` (see http://stackoverflow.com/questions/442404/retrieve-the-position-x-y-of-an-html-element) fixes the problem.

To reproduce issue, go to examples, turn on Device emulation in chrome devtools, apply some zoom to viewport and click on dropdown with container: 'body'. Dropdown is ill-positioned:

screen shot 2017-02-12 at 19 08 45

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -263,6 +263,12 @@
   var htmlEscape = createEscaper(escapeMap);
   var htmlUnescape = createEscaper(unescapeMap);
 
+  function getOffset(element) {
+    var docRect = document.documentElement.getBoundingClientRect(),
+        elemRect = element[0].getBoundingClientRect();
+    return {top: elemRect.top - docRect.top, left: elemRect.left - docRect.left};
+  }
+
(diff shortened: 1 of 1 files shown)


caseyjhol:
I've been playing around with as many settings as possible and I haven't been able to recreate this so far. This should be fixed when using Bootstrap 4, as Popper takes care of the positioning. I've been using the v1.12.4 bootstrap-select docs with Bootstrap 3 to test and haven't had any luck reproducing this (https://rawgit.com/snapappointments/bootstrap-select/cd8a9f5cd9eec8c57956649cb5db86dbd6f57d64/examples/index.html). Let me know if there's an easier way for me to recreate this issue.

NicolasCARPi:

PR #1645, opened 2017-02-21
Use tokens to store normalized text

andrews05:
This change is a simplification which eliminates the need for the "a*" pseudos by storing the normalized text as tokens. The actual behaviour is unchanged. It makes it easier to use your own search functions as you don't have to create two versions.

This also fixes a small error with the tokens where it would always put a null token - fixed by changing || to &&.


caseyjhol:
The search algorithm has been rewritten for the upcoming release.

Issue #1665, opened 2017-03-03
Performance issue loading page with large number of selectpickers with relatively large number of options


dw00001:
**Summary**:  I am building a page that contains a large number of select elements (~200) each with a large number of options (~500).  When using selectpicker for these elements the page has a very long load time and encounters the dreaded long running script warning on IE (performance is especially bad on IE).  I made a few tweaks to reduce the load times and have attached the test with the tweaked code.  Not sure if I have broken any functionality or overlooked any gotchas, but so far the changes seem to work correctly.  I wanted to provide these changes to see if some variation could be incorporated into the project to provide similar performance improvement. 

**Environment details**:  I encountered the issue in chrome, firefox and IE.  Performance is by far the worst in IE.  You can see the versions of jquery, bootstrap, and bootstrap-select in the attached example.

I tried to add appropriate comments in the code to explain the changes, but here is a rough summary of the changes:
1) changed code in render to only loop all options when select is multiple.  Otherwise just jump to the selected option using native JS.  Not a huge savings on chrome, but makes a bigger impact on IE.
2) added a lazyLoadLiElements option.  This option basically tells the selectpicker not to actually render the content on the dropdown menu when doing the inital load.  A bunch of expensive operations are skipped and then we bind an event so that when the dropdown menu is activated for the first time, we actually render its contents.  (it makes the the first time a little slower.  it is hardly noticable on chrome.  on IE it has a slight pause).  
3) As part of lazy load, I disabled the firing of the rendered event on initial loading.  This event is super expensive on IE. Looking at the code, this may cause an issue if an invalid option was already selected when the page loaded.  Perhaps this functionality can be implemented differently and you could provide an option to turn of this event for those that don't need it.  

Instructions for running the test:

1) extract contents of the attached zip file
2) load slow.htm in your browser
3) press "Create Select Fields(native)" button.  This will create 200 native select elements each with 500 options on the page
4) press "Selectpicker (orig)" button.  This will change the native selects into selectpickers using the original implementation and report the run time.

5) press reset (or refresh the page)
6) press "Create Select Fields(native)" button.  This will create 200 native select elements  each with 500 options on the page
7) press "Selectpicker(+perf changes)" button.  This will change the native selects into selectpickers using the performance tweaks described above and report the run time.

In my testing, I was seeing around an 80% improvement on chrome and around a 95% improvement on IE.
[...]

PR #1667, opened 2017-03-06
Lazy loading (extracted changes from provided archive, see issue #1665)

ghost:


anatoly-mytrucking:
For #1665

holtkamp:
Experienced the same performance reductions when large amount of `<select class="selectpicker"/>` are involved. The number of options (6) is not so large though.

anatoly-mytrucking:
@holtkamp It's a little bit "dirty" PR and I'm not sure about all changes. I just converted changes provided in #1665 into a commit so they can be easily inspected by other developers.

anatoly-mytrucking:
Did a proper extraction of changes. Previous time I did it on the top of upstream's master, now I used v1.12.2

Fixed couple of issues.

holtkamp:
Nice, I did some benchmarking with an e-commerce page that uses bootstrap select to allow to:
- indicate quantity of article
- indicate variations of article

For each article, 6 base variations are available. So for 1 article, 12 `<select/>` elements are used

I used this JavaScript file: https://github.com/MyTrucking/bootstrap-select/blob/2f16cb3a094c8b14844a8a10dd03874704abcdcc/js/bootstrap-select.js


FireFox
- 1 article, 12 select elements: ~130ms => ~130ms
- 10 articles, 120 select elements: 610ms => ~590ms
- 70 articles, 960 select elements: ~3800ms => ~2800ms

Chrome
- 1 article, 12 select elements: ~130ms => ~130ms
- 10 articles, 120 select elements: ~380ms => ~380ms
- 70 articles, 960 select elements: ~3800ms => ~2800ms

(I choose a maximum of 70 articles to test with since that is the maximum amount of articles displayed on a real-life page)

So there seems a small improvement but since the browser still "hangs" for around 3 seconds, I think other improvements are required. Maybe for Internet Explorer this PR is a life saver, but otherwise we should maybe consider postponing loading until an element is within the viewport of the browser...

anatoly-mytrucking:
@holtkamp just to be sure, have you added `data-lazy-load-li-elements="true"` to your `<select>` tags?

I think there's a room to improve this solution also, but I'm not sure if I will do it.
For postponing loading until an element is within the viewport of the browser you can try https://github.com/hakubo/Sloth and initialise bootstrap-select component using JS instead `data-` attributes.

holtkamp:
>just to be sure, have you added data-lazy-load-li-elements="true" to your <select> tags?

oops..., no I did not 😭 Guess its bed time 😄 . Just added `data-lazy-load-li-elements="true"` and this indeed changes the results significantly! Not such improvements as reported in https://github.com/silviomoreto/bootstrap-select/issues/1665#issue-211820788, but that may be due to the way of measuring (I use console.time() and just refresh a few times).

Chrome

1 article, 12 select elements: ~130ms => ~70ms
10 articles, 120 select elements: ~380ms => ~200ms
70 articles, 960 select elements: ~3800ms => ~1200ms

So +1 to merge!

holtkamp:
@anatoly-mytrucking, after receiving feedback from users, we decided to have dropdown menus actually dropdown when hovering the mouse over it, not only when clicking it. 

We realized this using: https://github.com/CWSpear/bootstrap-hover-dropdown

It seems this way of activating the menu, does not trigger the lazy loading of the `<li/>` items to kick in. Maybe there is another event than `click` to listen for here: https://github.com/silviomoreto/bootstrap-select/pull/1667/files#diff-ff31d80eb2d28b06752ee372aa065210R411 ?

Or any otherway to detect when the dropdown menu is activated. 

Clicking: activates dropdown menu AND triggers lazy loading
screen shot 2017-05-04 at 17 39 07

Hovering: activates dropdown menu but does not trigger lazy loading: empty menu
screen shot 2017-05-04 at 17 38 56

anatolykazantsev:
@holtkamp I switched from `show.bs.dropdown` event to `click.dropdown.data-api` one to get correct sequence of events when live search is on. But you can change it back to `show.bs.dropdown` if it works ok for you. Or maybe there's some `*.dropdown.data-api` for hovering.

holtkamp:
@anatolykazantsev ok, might be an idea to make this event configurable, defaulting to `click.dropdown.data-api`. But that would/should be a different PR. 

@silviomoreto what do you think? Can this be merged?

holtkamp:
@caseyjhol any idea on when a "final" `1.13.0` can be expected?

holtkamp:
@anatolykazantsev it seems https://github.com/silviomoreto/bootstrap-select/releases/tag/v1.13.0-alpha contains some performance improvements as well

frg:
Any update on this PR?

PR #1674, opened 2017-03-14
Fix <li> tags inside <option> being treated as options

gronostajo:
The `findLis()` function used a too broad selector which included `<li>` tags inside `<option>`. For example for this `<select>`:

```
<select>
  <option><ul><li>demo</li></option>
</select>
```

`findLis()` would find two `<li>`s in generated code: one for `<option>` and one inside it.

One side effect of this bug is that check mark is shown properly only for the first option. For following options the `selected` class is assigned to incorrect `<li>`.

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -709,7 +709,7 @@
     },
 
     findLis: function () {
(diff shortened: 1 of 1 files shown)


caseyjhol:
Available in v1.13.0-alpha.

Issue #767, opened 2014-10-17
Please update characters in accent insensitive search


deel77:
I am having problems using search without diacritics for some special characters in languages. I have made changes to your function and just want to recommend to use the fastest code from this example:
http://jsperf.com/diacritics/27
It removes much more diacritics characters than actual solution.

caseyjhol:
Released in [v1.13.4](https://github.com/snapappointments/bootstrap-select/releases/tag/v1.13.4)!

PR #1686, opened 2017-03-24
Improve diacritics normalization function (removing accents)

imbrish:
Similarly to a few other people I encountered problems when searching through a list that contained strings with diacritics. Other issues: #1551 #1587 #767.

Solved my problem using function from http://jsperf.com/diacritics/27 and decided to share the solution. 

Code is a little messy with the long mapping of characters. It probably could be simplified, but according to the benchmark from the link above this is the fastest method. However I haven't tested performance against the old function.


caseyjhol:
A more advanced normalizeToBase function will be used in the next release.

caseyjhol:
Released in [v1.13.4](https://github.com/snapappointments/bootstrap-select/releases/tag/v1.13.4)!

PR #1739, opened 2017-06-14
defaults-pl_PL.js

BooleanType:
Wrong settings keys selectAll and deselectAll. Should be selectAllText and deselectAllText.


BooleanType:
the same situation in defaults-pl_PL.min.js

PR #1749, opened 2017-06-24
for ajax

yy7054wyq5:
```javascript
this.$searchbox.on('input propertychange', function () {
          // change for ajax
 -        if (that.$lis.length === 0) {
 +        // use ajax must remove li
 +        // so that.$lis is null
 +        if (!that.$lis) {
            return;
          }
```


caseyjhol:
The ability to set options via an Ajax request is coming in an upcoming update. I've changed the way options are added which will make it easier to implement.

yy7054wyq5:
Glad to see it. Haha

PR #1762, opened 2017-07-11
add empty class to button when no value is selected

Cactusbone:
add empty class to button when no value is selected, allowing for css customization when empty (or not), for example to style placeholder


caseyjhol:
The class `.bs-placeholder` is already added when there is nothing selected.

Cactusbone:
nice ! I had not noticed it :) thanks

PR #1763, opened 2017-07-11
Html tag option

Cactusbone:
html tag option, to decide between using `span` or `i`


caseyjhol:
What are the issues with using span?

Cactusbone:
I'll check locally, since we're using material design icons, all the examples uses `i` but i'll check if default styles really uses it.

Cactusbone:
we're migrating from 1.5.4 to the latest version so keeping `i` was easier during migration :)

Cactusbone:
i cleaned up our css to not use i . i'll close this pull request :)

PR #1814, opened 2017-09-27
Actualize 'selected' attribute on Select All action.

igoodwin:
If some option not selected before calling 'selectAll' action, click on the some option not working properly. It detects actual selected items by calling jQuery this.$element.val() method which ignore 'selected' property state.


caseyjhol:
Setting the `selected` attribute is only necessary when pre-selecting options on page load. I'd prefer not to have to set it unless necessary, as it could be a potential performance hit. If you can elaborate on the advantages of this PR I may consider it.

Issue #1803, opened 2017-09-08
Prevent LiveSearch textfield from getting focus (open keyboard on mobile)


GeeZuZz:
When on mobile, tapping on the picker (with liveSearch=true) will open the keyboard, which takes some time to show up, and covers half the screen, making it hard to pick items. Is it possible to prevent the keyboard from showing up (prevent focus to the textfield?), unless the field is specifically clicked on?

scheMeZa:
bump. same problem.

himdel:
Not only problematic on mobile:

when selecting items from a multiselect with more than a page of items,
clicking each item refocuses the search box, making that item to scroll out of view.

So, it is impossible to select multiple items after each other without scrolling again and again, unless the items are near the top of the list.

caseyjhol:
@himdel Can you post an example of this issue? What version of bootsrap-select are you using? That is not an issue I've encountered before.

himdel:
Sure, I'm on 1.12.4... adding a screengrab:

 
![20181121123244_magic](https://user-images.githubusercontent.com/289743/48841882-abe54800-ed8a-11e8-8848-c8ec6c443c29.gif)

kriit24:
![screenshot_20190211-171051_chrome](https://user-images.githubusercontent.com/1489085/52572294-2ff61b80-2e20-11e9-94b5-c71b95f4018d.jpg)

I have problem that bootstrap select moves up in mobile.
Latest bootstrap and bootstrap-select version

Issue #1511, opened 2016-09-14
Option to not focus live search when opened


ProfessionalNihilist:
Hello,

I would like to be able to include a search field but not have it focused when the control is opened. This is quite annoying on devices with onscreen keyboards where we do not want to go back to just using the native control.

I can't see if this is possible from reading the documentation or a quick look over the code.

Regards

nithin-ideas2it:
+1

dipeshcct:
@caseyjhol can you please tell me where is solution for  this?

amitsquare:
Hi Guys ,
After looking at the source code .
These lines gives focus to searchBox in the case of Live Search..
 [Source Lines for Focus](https://github.com/silviomoreto/bootstrap-select/blob/master/js/bootstrap-select.js#L1408-L1410)

Either you can comment it out for permanent effect .you can pass additional option to recognize & then disable it .
Hope this will help you out .

Let me know

RALGIE:
I have the same problem. When I'm on an IPAD and I select a value in the list the search field is focused and the onscreen keyboard pops open. This is very annoying. 

can you remove this behavior?

lorsulic:
Will this solution ever be merged into the project? Downloading and manually editing the source file isn't a good approach.

a-vasyukov:
````js
/* Hack for focus blur from liveSearch input */

var blurState = false;

$('.bs-searchbox .form-control').on('focus', function() {
    if (!blurState) {
        $(this).blur();
        blurState = true;
    }
});

$('.selectpicker').on('hide.bs.select', function() {
    blurState = false;
});
````

heyitsedward:
> I have the same problem. When I'm on an IPAD and I select a value in the list the search field is focused and the onscreen keyboard pops open. This is very annoying.
> 
> can you remove this behavior?

Did you find a fix for this? I'm having the same issue.

SewookJung:
Just Fixed the bootstrap-select.js file directly.

1. Find the setFocus funtion 
2. Fixed the code
<pre class="javascript" style="font-family:monospace;"><span style="color: #000066; font-weight: bold;">function</span> setFocus<span style="color: #009900;">&#40;</span><span style="color: #009900;">&#41;</span> <span style="color: #009900;">&#123;</span>
[...]

PR #1815, opened 2017-09-27
Fix keyboard opening problem on click items on mobile

scriptype:
This pull request solves the problem mentioned in the issues: #1803 #1511.

Problem explanation:
 - Enable `live-search` and `multiple`
 - Open the page in a mobile device
 - Click an item from the list
 - Keyboard shows up, this is annoying and bad UX.

I've made a tiny addition to the condition for focusing to `searchBox` inside onClick event handler of list items. It now also checks the `mobile` option before focusing the input. If `mobile` option is present, it won't focus, thus preventing the keyboard opening issue.

I couldn't reproduce the issue in an isolated environment, but I've added a related test-case into `test.html`. You can copy the test-case I added to `test.html` into current `test.html` to see the problem occurring, and then check the example in my branch to see it's fixed.


ricdecastro2009:
Were you able to display this componte through the Iphone mobile?
I'm having problems with the default behavior of the <select> tag and this component in the mobile
Can you help me?

NicolasCARPi:

PR #1818, opened 2017-09-28
data tokens not updating on selectpicker refresh fix

tstulka:
Feeding selectpicker with new data caused data-tokens to remain from the previous one. This was probably caused by a bug in jQuery .data() method. I've replaced this one to .attr() and everything worked like a charm.


caseyjhol:
Not compatible with the new method used to generate options.

PR #1833, opened 2017-10-12
Add jsDelivr hits badge

LukasDrgon:
Hey, I noticed that bootstrap-select gets 12M hits per month on jsDelivr, so I thought you might like this badge.


caseyjhol:
Added in latest release.

Issue #1624, opened 2017-02-03
Multiple select title showing line breaks


kpagcha:
I am using a multiple select and the `title` attribute of the `select` tag is displaying the data separeated by line breaks.

![1](https://cloud.githubusercontent.com/assets/3865095/22585640/05b9737a-e9f9-11e6-867e-10b4a0a3ab4a.png)

aknosis:
Did you find a resolution to this issue? I am seeing it myself. I'm guessing the issue is that it is taking the <option> content verbatim and therefore including the white space.

e.g.

```html
<option>
    Test
</option>
<option>
    Test 2
</option>
```
vs.
```html
<option>Test</option>
<option>Test2</option>
```

caseyjhol:
Fixed in v1.13.0. See #2063.

PR #1851, opened 2017-11-03
updates building of title attribute to trim option label white space

aknosis:
This solves an issue where the title attribute is including white space that is inside <option> tags when multiple options are selected.

Given html source like this:

```html
  <select class="selectpicker" multiple>
      <option>
        Apple
      </option>
      <option>
        Orange
      </option>
      <option>Corn</option>
      <option>Carrot</option>
  </select>
```
Select Apple and you end up with: `title="Apple"`.

If however you select Apple and Orange you end up with:
```html
title="Apple
                                        , 
                                            Orange"
```

Only the <option> with inner white space have this issue, selecting Corn or Carrot do not change anything. e.g.

If however you select everything you end up with:
```html
title="Apple
                                        , 
                                            Orange, Corn, Carrot"
```

This was noted previously in #1624, reproduced with v1.12.4 using sample html here.

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -1179,7 +1179,7 @@
 
       //Fixes issue in IE10 occurring when no default option is selected and at least one option is disabled
       //Convert all the values into a comma delimited string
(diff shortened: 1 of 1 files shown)


caseyjhol:
Fixed in v1.13.0. See #2063.

PR #1857, opened 2017-11-13
Add regex matching as a live search style

MSanteler:
Adds a new `liveSearchStyle` called `regularExpression` which preforms a match-check using regex and ignoring whitespace. This allow for slightly fuzzier finding such as entering "Penn State", "Pennsylvania State University (The)"

Open to feedback on the the expected value (`regularExpression`) since this is a bit more specific than a generic regex match.

With credit to @robadams


caseyjhol:
I like the idea. The implementation would have to be slightly different with the way search is implemented in v1.13.0+. Also, it would be great if it had the option to provide a custom regex.

NicolasCARPi:

PR #1873, opened 2017-12-07
Added Option to not close dropdown on select

samklonaris:
Add option `closeOnSelect`, which defaults to true.  If set to false, the dropdown will not close when you select an option.


NicolasCARPi:

PR #1898, opened 2018-01-16
Added ability to provide custom item text when count format is provided

jeffbeagley:
We wanted to change the word "items" when multiple items are selected. So I defined an option to provide it. This is one of my first github contributions so take it easy hah.

### DIFF
--- a/dist/js/bootstrap-select.js
+++ b/dist/js/bootstrap-select.js
@@ -343,8 +343,9 @@
   Selectpicker.DEFAULTS = {
     noneSelectedText: 'Nothing selected',
     noneResultsText: 'No results matched {0}',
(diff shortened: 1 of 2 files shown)


ManuZenou:
will this be merged any time soon?

caseyjhol:
Thanks for your contribution! Seems kind of overkill to add a new option just to replace 1 word. You would have the same outcome if you just set `data-count-selected-text: '{0} things selected'`. Admittedly, that won't work if `selectedTextFormat` is `count > 0` (you'll have to use a custom function instead), but this PR won't address that issue either. I think a better option, which I'll try to add in one of the next updates, would be to add support for an array, where the first item is the text for when 1 item is selected, and the second is for multiple. That would allow easier customization when using data attributes.

PR #1933, opened 2018-03-02
bugfix - "Nothing Selected" with 1 item selected with data-selected-text-format

jeffbeagley:
…a-selected-text-format was provided, the selectbox would say "nothing selected" when one item was.

### DIFF
--- a/dist/js/bootstrap-select.js
+++ b/dist/js/bootstrap-select.js
@@ -1211,7 +1211,7 @@
       if (this.multiple && this.options.selectedTextFormat.indexOf('count') > -1) {
         var max = this.options.selectedTextFormat.split('>');
 
-        if ((max.length > 1 && selectedItems.length > max[1]) || (max.length === 1 && selectedItems.length >= 2)) {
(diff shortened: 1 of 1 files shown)


jeffbeagley:
This pull request conflicts with using count > 1, back to the drawing board!

caseyjhol:
Can you provide an example of the bug?

jeffbeagley:
Here's a fiddle. You can see with 1 item selected, and option "count" provided, it still states nothing selected.

https://jsfiddle.net/jeffbeagley/DTcHh/51050/

caseyjhol:
Ah - I see. This is only happening with v1.13.0-beta. Should be an easy fix.

caseyjhol:
Fixed here: https://github.com/snapappointments/bootstrap-select/commit/41ebe2220e82279200937c51ef5c8ffb90fe2637. Available in the v1.13.0-dev branch in that repo. Will be available in v1.13.0 or v1.13.0-beta.1.

caseyjhol:
Released in [v1.13.0](https://github.com/snapappointments/bootstrap-select/releases/tag/v1.13.0)!

PR #1974, opened 2018-05-16
Correctly looking for parent data, using .closest instead of parent()......

### LINKED ISSUE #1085: Cannot read property 'options' of undefined
GREAT PLUGIN!  Thanks for the hard work.

A javascript error is thrown when combining `data-live-search="true"` with option groups

In _boostrap-select.js_ (line 1363):  

`if (that.options.liveSearch) $parent = $this.parent().parent();`

Example: 

```
<select id="sel_assign"
    class="selectpicker"
    data-width="100%"
    data-size="auto"
    data-live-search="true"
    title="Select an Assignee">
    <optgroup label="Crew Members">
    <option value="1">Smith</option>
    </optgroup>
</select>
```

Help!

hems:
Fix buggy workaround to find the wanted parent data, fixed #1085 at least for me.

I'm really not sure if it's a good idea to have this selector ( .dropdown-menu ) hardcoded here because we didn't spend time with anything other than solving this bug on our specific application.

Please fix the selector if needed.

Thanks for the library.

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -1659,6 +1659,11 @@
           };
 
 
+      // we destroy the modal/popup using the ESC key and then turns out we get
+      // here and the parent doesn't even exist anymore, in this case fail silently.
+      // probably not the best idea, but works.
+      if(!$parent.length) return false;
+	    
       isActive = that.$newElement.hasClass('open');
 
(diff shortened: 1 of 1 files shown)


TamizhJayapal:
Calling click on an option element inside a select doesn't update the model in bootstrap select
can u help me?

TamizhJayapal:
Calling click on an option element inside a select doesn't update the model in bootstrap select.
can u help me?@silviomoreto

silviomoreto:
Is it relatted to this PR?
If not, @TamizhJayapal please open an issue regarding that.

PR #1991, opened 2018-06-19
Feature aria label nvda support

tstulka:
Currenlty bootstrap-select is not supporting NVDA software properly (NVDA can't read elements on the list), I've added an aria-label parameter to fix this issue.


caseyjhol:
Please follow #2153.

PR #2133, opened 2018-11-01
Rename defaults-ua_UA.js to defaults-uk_UA.js

LorenzKeel:
ISO 639-1:2002, Codes for the representation of names of languages — Part 1: Alpha-2 code


caseyjhol:
Addresses #2338. For now, we'll have to copy the file with the new name to prevent any breaking changes. Then, the old files can be removed in v2.0.0.

PR #2154, opened 2018-11-22
Add liveSearchFocus option - only focus live search when searching if the setting is false

### LINKED ISSUE #1511: Option to not focus live search when opened

### LINKED ISSUE #1803: Prevent LiveSearch textfield from getting focus (open keyboard on mobile)

himdel:
This adds a new setting - `liveSearchFocus`
  defaults to `true`

When `true`, no change in behaviour
When `false`, we never focus on live search, except when handling a keydown event (so starting a search?)

I believe this:

Fixes #1511 - it adds the requested option
Fixes #1803 - adding `data-live-search-focus="false"` when on mobile should be enough. (mobile detection itself should not be handled by bootstrap-select IMO)
Closes #1815 - essentially the same change

Cc @caseyjhol .. if everything is OK, I'd like to request a 1.12.* release with this change backported please
(Or should I create a second PR against `v1.12.4`?)

---

gif of the problem this fixes: https://github.com/snapappointments/bootstrap-select/issues/1803#issuecomment-440648934


himdel:
Ping @caseyjhol .. would you have time to review please?

derigel23:
@caseyjhol is the 1.12.x release coming? Especially with this fix?

himdel:
@derigel23 I needed the release soon, so in the mean time, I released this as `@pf3/select`.

It's forked from 1.12, I'm not really planning to do any active maintenance there, but as long as ManageIQ is depending on bootstrap-select 1.12, the fork will stay alive and I can merge non-invasive patches :).

(One thing that will probably appear in that fork soon is I plan to remove the AMD/UMD/CJS... wrapper to make it easier to run through webpack - meaning don't rely on the package supporting IE without precompilation, but you should be fine otherwise.)

caseyjhol:
@himdel Hoping to get this implemented soon. However, I would love to know _why_ you are unable to use v1.13.5, as there should not be any breaking changes. If there are any breaking changes, if you could tell me what they are so I can get them fixed it would be much appreciated.

himdel:
Oh, of course.. I haven't really investigated into *much* depth there.

But the short answer is, we're using http://patternfly.org (a businessy bootstrap 3 clone with themes and a bunch of libraries bundled in), and:

a) it depends on bootstrap-select 1.12
b) updating bootstrap-select to 1.13 breaks styling of those selects

IMO it's probably a bug on the patternfly side, but unfortunately they also created a new (IE-incompatible) version (patternfly 4), so there is not so much support for the old ones these days.

caseyjhol:
@himdel Any chance you could send me a screenshot, or a preview of the behavior when using v1.13.5 (feel free to email me)? If 1.12 works, 1.13 should be working as well (otherwise somewhere a breaking change has been introduced). I'm not opposed to releasing a new v1.12.x, but I'd prefer to get to the source of the problem.

heyitsedward:
@caseyjhol any idea when we could see this implemented?

himdel:
@caseyjhol This is how it looks when combining patternfly and bootstrap-select 1.13:

bad

I'll try to get a jsbin/... working..
EDIT: here is a static (js-less) version of the page https://www.himdel.eu/bootstrap-select.html

derigel23:
@caseyjhol for my case, https://pogoquest.team23.ru/
"data-width": "110%" option is not working.
But, even this bug aside there is no data-live-search-focus option in 1.13.5 and live search focus problem exists.
P.S. Oh sorry, misread your comment. bootstrap-select trunk doesn't contain fix yet. But my issue with 1.13.5 is described.

caseyjhol:
@himdel Thanks for the screenshot/demo. So if I'm understanding correctly, you're unable to use v1.13.5 CSS?

himdel:
@caseyjhol That's right, yes. (The javascript bits are probably fine, though I haven't really tested that, given the css problems.)

fredriksundstrom:
Hey @caseyjhol, has this type of option been implemented elsewhere, was there a reason for not implementing it, and will something like this be added soon? I'm working the UX on a webapp and we use this great addon. This change makes sense, though, since the behavior on mobile is very "jumpy" with the keyboard popping up, re-positioning of the dropdown etc.

NicolasCARPi:

Issue #1832, opened 2017-10-11
bootstrap-select compatibility with Jquery version 3 or more


maheshparadkar:
I am trying bootstrap select with Jquery latest version 3 it seems its not working properly is there any solution for that or release new version for that

bartclaeys:
I confirm that is does work perfectly with jQuery 3x.

beerlander:
Even the latest version of bootstrap-select is still using fn.change() and fn.click() event shorthands in its js that are deprecated in jQuery 3.x (enabling jQuery migrate plugin will point that out).

mattfeury:
yup these functions (`.click` and `.change`) are deprecated and should probably be replaced
https://github.com/jquery/jquery-migrate/blob/master/warnings.md#jqmigrate-jqueryfnclick-event-shorthand-is-deprecated

garima2094:
I also tried bootstrap select with jquery3, and the Bootstrap select plugin is not working properly. is there any solution available?

caseyjhol:
Released in [v1.13.6](https://github.com/snapappointments/bootstrap-select/releases/tag/v1.13.6)!

PR #2173, opened 2019-01-10
Fix for jQuery 3x

emealcubo:
Just a few changes so it can work with jQuery 3x without the migrate.

Fix for this issue: #1832


caseyjhol:
Looks good - thanks!

PR #2175, opened 2019-01-11
Disable the button when disabled

Aaron-P:
Disable the button that replaces the select input when the select input is disabled and vice versa.


NicolasCARPi (changes requested):
Could you fix conflicts or redo one based on v1.14-dev branch please?

PR #2176, opened 2019-01-11
Validaton selector

Aaron-P:
The .selectpicker class may not exist on the select element if it's initialized directly via javascript, which breaks the validation styling.  Changing the selectors to look for the select element directly fixes it.


caseyjhol:
Released in [v1.13.12](https://github.com/snapappointments/bootstrap-select/releases/tag/v1.13.12)!

Issue #2153, opened 2018-11-21
 Screen readers doesn't read elements in expanded list (NVDA and JAWS).


mpawelski:
Screen readers doesn't read elements in expanded list (NVDA and JAWS).

Steps to reproduce:
1. Got to [demos page](https://developer.snapappointments.com/bootstrap-select/examples)
2. Run NVDA (to download from [here ](https://www.nvaccess.org/download/))
3. Try to focus on first bootstrap select example with keyboard (use Tab and arrows).  You will hear something like this when you will try to exand list _"mustard button collapsed, expanded, combo box collapsed, list expanded"_. Now you have expanded list but changing selected list item with arrows or tab key doesn't make it read by screen reader (you have to use mouse to hover over this items to read it).

I was trying to find the solution for it but I'm not that familiar with codebase. It seams that if I remove two roles you set for divs outside of `<ul>`("combobox" and "listbox") screen reader starts to read this items. But it break "focus trap" on expanded list and seems to break other stuff (list is not always expanded). Small changes I did to investigate it are [here](https://github.com/mpawelski/bootstrap_select_screen_reader_tests/commit/b2c8845b64756dde126026068db7ff026f234221)

Anyway, that's for working on this library!

sanjuroo:
everyone ignore this basic issue?

caseyjhol:
Note to self: see https://w3c.github.io/aria-practices/examples/listbox/listbox-collapsible.html.

AdrianHostettler:
Hy, just wanted to let you know, that someone already did some kind of fix for it. Not sure if standards of w3c specs are already met, but it would work for NVDA Screenreader, that's what i tested on v1.13.5.

this is the pull request that solved the issue (but maybe only temporary? the fix worked on v1.13.5.)
https://github.com/snapappointments/bootstrap-select/pull/1991

May i ask on which release you are targeting to implement this fix? Already in v1.x.x version, or targeting v2.x.x version?

With kind regards, and thx already for your huge work so far on this library!

Bene98:
Hey, I had the same problem with chromevox and for me works this solution: #2255

caseyjhol:
Hoping to have this available in the next release, as it's a bug.

caseyjhol:
Released in [v1.13.10](https://github.com/snapappointments/bootstrap-select/releases/tag/v1.13.10)!

PR #2255, opened 2019-04-09
fix focus in dropdowns on keyboard navigation

Bene98:
At the moment screenreaders don´t work on the dropdowns, because the focus is not set correctly.


caseyjhol:
The focus is set correctly, but `aria-activedescendant` isn't being utilized. Please follow #2153.

PR #2257, opened 2019-04-09
Fix displayed checkmarks when setting the values programatically with…

### LINKED ISSUE #2256: selected option styling is not updated if .selectpicker('val', x) is called while dropdown menu is open
Reproduce:
Select some options in the dropdown. In on-change event call .selectpicker('val', someOptionValue). The checkmarks are not updated.

rbonomo:
… the menu displayed.

Closes #2256

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -2567,6 +2567,16 @@
 
         changedArguments = [null, null, prevValue];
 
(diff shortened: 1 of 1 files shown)


caseyjhol:
Thanks for the PR! I ended up fixing this in 8d5d51f with a different approach.

Issue #2263, opened 2019-04-23
Selecting non-existent option throws error in v1.13.10


Stoneguard001:
The system assigns the previous selected option to a variable, then sets selected to false. But if that value is -1, an undefined variable causes the code to crash.

This occurred anywhere I was using a data-live-search=true value and there was no default selection value. The fix is to simply check for undefined on the variable

```
if (!that.multiple) { // Deselect all others if not multi select box
  if (prevOption !== undefined) {
     prevOption.selected = false;
  }
```

caseyjhol:
It shouldn't ever be possible for a single select to not have a selected option (i.e. selectedIndex should never be -1). That being said, what should be possible, and what is actually possible are often different. Could you provide me with an example? Or let me know the browser you are encountering this issue with?

ghost:
Hi, Don't know if this helps but I'm having the same issue when using version 1.13.10 but I DONT have the issue when using 1.13.9. I'm not using data-live-search but it's happening in IE11 & Chrome on any single select list with nothing selected.

Stoneguard001:
I tried working with this last night on a jsfiddle but could not re-create it. I will check what version of bootsrtap-select it is using and see what I can do to duplicate it

felixvictor:
I have the same issue. My unchanged code doesn't work with 1.13.10 but does work with 1.13.9.

giero:
Same problem :(

Here's an example: https://jsfiddle.net/bp9hy1sz/

Stoneguard001:
Thank you! Not sure why I could not get my jsfiddle to work.

PhilPJL:
Same issue here.  Works with 1.13.9, not with 1.13.10.  Setting data-title is a work around.

vandetho:
Another solution would be to add an empty option tag with selected attribute.

caseyjhol:
Another example:

> In Version 1.13.10
> 
> Here the bug:
> 
> When you select a non existing option by code in select, it breaks the selection by hand:
> 
> Here a [code pen](https://codepen.io/anon/pen/WqYama?editors=1111) of the issue.
> 
> Here the v1.13.9 [code pen](https://codepen.io/anon/pen/YoRdVM?editors=1111) working

caseyjhol:
Released in [v1.13.11](https://github.com/snapappointments/bootstrap-select/releases/tag/v1.13.11)!

PR #2264, opened 2019-04-23
Fix for unselected live search value (#2263)

Stoneguard001:
Simple fix for the unselected live search value not having a selectedIndex.


caseyjhol:
Released in [v1.13.11](https://github.com/snapappointments/bootstrap-select/releases/tag/v1.13.11)!

PR #2354, opened 2019-11-17
added missing semicolons

Ciki:


NicolasCARPi:
Thank you for wanting to contribute, but a better fix is https://github.com/snapappointments/bootstrap-select/pull/2678 ;)

PR #2361, opened 2019-12-04
Issue "Cannot read property 'findLis' of undefined"

przemyslaw-szustak:
When bootstrap-select was created dynamically on modal window and I close it then error occurred:
'Cannot read property 'findLis' of undefined'. 

This code prevent for that.


caseyjhol:
Please provide an example of the issue.

Issue #2147, opened 2018-11-13
Add placeholder option / deprecate title option


caseyjhol:
A new option - `placeholder` will function the same as `title` does currently. `placeholder` will be available and `title` will be deprecated in the next minor release.

`title` will be removed in the next major release. The `title` option will then behave as it would be expected (a tooltip with the title text will be displayed when hovered over).

Addresses #977.

caseyjhol:
This change will be a part of the v1.14.0 release. This issue will be updated after it's made available.

caseyjhol:
Released in [v1.14.0-beta](https://github.com/snapappointments/bootstrap-select/releases/tag/v1.14.0-beta)!

vukasinpetrovic:
Hi @caseyjhol,
I tried adding placeholder to option elements, but it is not displayed. Instead whole text of options is shown when it's selected. I'm using regular single select, not the multiple one.

Issue #1416, opened 2016-06-14
Update the title


bonsall:
It looks like i can't update the title without destroying the selectpicker and changing the html, like this.

```
$('.selectpicker').selectpicker('destroy');
$('select.selectpicker').html(someOptions);
$('.selectpicker').selectpicker({title = 'New Title'});
```

Did I miss something along the way? Can this be added as a new feature?

caseyjhol:
This change will be a part of the v1.14.0 release. This issue will be updated after it's made available.

caseyjhol:
Released in [v1.14.0-beta](https://github.com/snapappointments/bootstrap-select/releases/tag/v1.14.0-beta)!

PR #2369, opened 2020-01-06
Add placeholder option / deprecate title option (#2147)

hndcrftd:
Title option is deprecated (but still works for backwards compatibility). Placeholder option is now used.
'placeholder' can be set as html attribute, data attribute or set in the options object on init.
Updates to html attr or data attr will reflect upon calling refresh method.
Title, if set, is appended as title attribute on $button.
Title usage as placeholder can be removed by deleting line # 133.

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -64,6 +64,8 @@
    */
   var DATA_URL_PATTERN = /^data:(?:image\/(?:bmp|gif|jpeg|jpg|png|tiff|webp)|video\/(?:mpeg|mp4|ogg|webm)|audio\/(?:mp3|oga|ogg|opus));base64,[a-z0-9+/]+=*$/i;
 
+  var ParseableAttributes = ['title', 'placeholder']; // attributes to use as settings, can add others in the future
+
   function allowedAttribute (attr, allowedAttributeList) {
     var attrName = attr.nodeName.toLowerCase()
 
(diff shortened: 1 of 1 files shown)


caseyjhol:
Thanks for this! Reviewing. Would you mind changing this to the v1.14-dev branch instead? https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/changing-the-base-branch-of-a-pull-request.

Addresses #2147 and #1416.

caseyjhol:
With these changes, it's no longer possible to set the title programmatically.

v1.13.14: https://plnkr.co/edit/UBDl2ikH9aMwSq0l
5bb8f61: https://plnkr.co/edit/tfPLd02FWQJ3s3Qd

caseyjhol (changes requested):
With these changes, it's no longer possible to set the title programmatically.

v1.13.14: https://plnkr.co/edit/UBDl2ikH9aMwSq0l
5bb8f61: https://plnkr.co/edit/tfPLd02FWQJ3s3Qd

caseyjhol:
Figured out the issue. Going to merge this and fix issues after.

jrichview:
It looks like I will need this. My organization is pushing back on it displaying title as the placeholder. They only want that to be a normal tooltip behavior.  I need to be able to set the placeholder to show the selected value or "-Select-" if nothing is selected.  The "noneSelectedText" is being set but it is being overridden by the title attribute.

Is this in a released version yet?

PR #2392, opened 2020-02-12
Bug/2391 set style with document fragment parent

### LINKED ISSUE #2391: setStyle fails if newElement.parentNode is a document fragment
With version 1.13.12 and Bootstrap 3, `setStyle` fails if `newElement.parentNode` is a document fragment because `classList` is undefined in this case. This used to not be a problem with version 1.13.5.

Also, the call to `buildData()` in `refresh()` should come before `render()` like it is done in `init()`.

Mobe91:
fixes #2391

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -2969,9 +2969,9 @@
       this.options = config;
 
       this.checkDisabled();
+      this.buildData();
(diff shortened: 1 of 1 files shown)


caseyjhol:
Thanks! Will be in the next release.

PR #2394, opened 2020-02-14
[#2393] consider custom content for li height computation

### LINKED ISSUE #2393: liHeight does not consider custom option content when computing li height
Currently, liHeight creates an artificial `li` element with static content to compute its `offsetHeight`. This height is later used for virtual scrolling.
However, when a user supplies custom content via the `data-content` attribute, the current approach does not yield the correct li height.

Mobe91:
fixes #2393


caseyjhol:
Thanks for this! Had to make a few changes to account for cases where the first element isn't an option. This was added in c736ea9.

PR #2396, opened 2020-02-18
[#2395] for virtual scroll, show last chunk in case of no previous match

### LINKED ISSUE #2395: virtual scroll sometimes jumps to first chunk instead of showing last chunk
Due to imprecisions in scrollTop vs data position computation, the computation of the currentChunk does not yield a match in some cases. The fallback is to show the first chunk. This results in an empty menu because the current scrollTop and the first chunk margins do not match in this case.

Mobe91:
fixes #2395

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -1212,7 +1212,7 @@
 
           if (!size) break;
 
-          if (currentChunk === undefined && scrollTop - 1 <= that.selectpicker.current.data[endOfChunk - 1].position - that.sizeInfo.menuInnerHeight) {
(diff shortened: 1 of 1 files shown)


BT-aestebanez (approved):

caseyjhol:
Thanks for the PR. Could you link me to a demo of the problem?

Mobe91:
As far as I can recall, this problem only happened with #2394 applied. However, I just tested with 1.13.14 + #2394 and it seems like this PR (#2396) is no longer needed. Let me do some more testing on Monday. If nothing shows up I will close this PR.

Issue #2367, opened 2019-12-30
countSelectedText not working when only 1 item is selected


cbrianball:
I have tried setting `countSelectedText` to both a string template as well as a function.
I've also tried setting `selectedTextFormat` to both `count` and `count > 0`.

It works fine if I select more than 1 item, but if I only select 1, I just see that item's text (which in my case doesn't look good since I use a template for each item).

I can understand the default of `selectedTextFormat: "count"` only showing the `countSelectedText` value if more than 1 is selected, but at the very least, I would expect "count>0" to work where it shows my `countSelectedText` template if only one item is selected.

I've narrowed down the issue to one line of code:
https://github.com/snapappointments/bootstrap-select/blob/2c39162d5e4f6584350e14cfdfe3811f27d50afd/js/bootstrap-select.js#L1664

Here's a quick reproduction of the issue (I kept it as simple as possible, even opting to use the default `countSelectText` (and not defining my own).

http://jsfiddle.net/k2ph0vcm/

mikemix:
Still doesn't work.

mirko77:
I have the same issue

NicolasCARPi:
Fixed in beta3 version.

PR #2403, opened 2020-02-26
Start counting selected options at 1

nbdavies:
This PR addresses issue #2367 : currently if set `selectedTextFormat` to `count` or `count > 0`, when the user selects one option, the text displayed can be determined using `countSelectedText` unless the user happens to select only one option, and then that one option is displayed. That can look very inconsistent, especially if the `<select>` has a limited width and the selected option is longer.

The change here just fixes the `selectedCount` threshold at which the `countSelectedText` function comes into play (from 2 to 1). Test cases are included. I don't think the documentation needs to be updated, since that's already accurate:

> [countSelectedText] Sets the format for the text displayed when selectedTextFormat is count or count > #. {0} is the selected amount. {1} is total available for selection.

> [selectedTextFormat] 'count' displays the total number of selected options. 'count > x' behaves like 'values' until the number of selected options is greater than x; after that, it behaves like 'count'.

This _could_ be considered a breaking change to anyone who's expecting/relying on the previous behavior, but it's easy to get back the previous behavior by setting `selectedTextFormat` to `count > 1`

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -1673,12 +1673,12 @@
       if (this.options.selectedTextFormat === 'static') {
         titleFragment = generateOption.text.call(this, { text: this.options.title }, true);
       } else {
-        showCount = this.multiple && this.options.selectedTextFormat.indexOf('count') !== -1 && selectedCount > 1;
+        showCount = this.multiple && this.options.selectedTextFormat.indexOf('count') !== -1 && selectedCount > 0;
 
         // determine if the number of selected options will be shown (showCount === true)
         if (showCount) {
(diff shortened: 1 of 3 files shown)


mikemix:
Are we merging this?/

caseyjhol:
Ideally, `count` behavior would remain unchanged and only support for `count > 0`  would be added. This would reduce the chances of this breaking someone's setup.

Edit: support for `count > x` has been added here: https://github.com/snapappointments/bootstrap-select/commit/036d0db12b01f44d42b469fe9be0554ef21e44bb and will be available in the next release after v1.14.0-beta2.

PR #2424, opened 2020-04-06
Delete spaces before applying a search

yousfiSaad:
In case that the shown list contains spaces before the actual string, the startwith search will not work
This is a real case that I found when using this library


NicolasCARPi:
@caseyjhol this looks like a mergeable PR to me.

PR #2506, opened 2020-07-24
Work with React

melvinotieno:
Since version 16.0 of React, React.createClass has been deprecated and moved to a separate package


caseyjhol:
Did you mean to submit this to https://github.com/tjwebb/react-bootstrap-select?

PR #2540, opened 2020-11-06
fixed broken rendering on Safari new iOS 14.2

depontimatteo:
Pull request for the issue opened this morning, here:

https://github.com/snapappointments/bootstrap-select/issues/2539

Thank you, btw

Keep safe and healthy

Teo


andrerom:
Was this fixed in later iOS versions? I'm on 14.4 and can't see any issues on demo/doc page

PR #2576, opened 2021-02-26
better jquery selectors on loooong lists

rogeliotg:
jquery selectors are somehow optimized on Chrome but not on Edge.
Trying to find into 100000 options list dies on Edge, so I propose use children instead


caseyjhol:
This doesn't work if the first option is inside an optgroup. How does this compare?

```js
firstOption = this.$element[0].options[0]
```

Issue #2578, opened 2021-02-26
Using "allowClear" doesn't update the value


joelstein:
The new "allowClear" button doesn't update the value, though it does update the visual select element.

Environment details: Mac OS 11.2.1, Brave 1.20.108, jQuery 3.5.1, Bootstrap 4.6.0, bootstrap-select 1.14.0-beta.

Example which shows bug in action: https://plnkr.co/edit/XDJrlaaDguYig2sl

Related: #1449

caseyjhol:
The value _is_ actually updating, but it's not triggering the change event.

joelstein:
Ah, thanks, I wasn’t sure. I put together a pull request which seems to work for me, but I didn’t fully understand the code...

joelstein:
Perfecto! Thank you so much.

caseyjhol:
Released in [v1.14.0-beta2](https://github.com/snapappointments/bootstrap-select/releases/tag/v1.14.0-beta2)!

PR #2579, opened 2021-02-26
Fix allowClear so that it updates the value

joelstein:
See #2578.


caseyjhol:
This looks good, but I think we may want to set `changedArguments` as well. I'll need to investigate a bit.

PR #2585, opened 2021-03-12
Fixes two `TypeError` exceptions when up/down arrow pressed

rjregenold:
It looks like issue #2035 is happening again in the latest v1.13 release. When using live search, if a user enters a query that has no matches and then presses the up or down arrows, two different `TypeError` exceptions can occur. This PR fixes both exceptions.

For reference, here are the steps to reproduce the issue, copied from the original issue:

Browser: All
OS environment: All

Steps to reproduce:

1. Open selctbox and type "xxx" to search
1. No results matched "xxx" will appear
1. Press DOWN key
1. Error is thrown

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -2913,7 +2913,7 @@
 
         liActive = that.selectpicker.current.elements[liActiveIndex];
 
-        that.activeIndex = that.selectpicker.current.data[liActiveIndex].index;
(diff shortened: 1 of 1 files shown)


caseyjhol:
~~I can confirm that this issue exists, but this will require a different fix for v1.14.0, as it's been refactored.~~

This fix should work in v1.14.0 as well.

PR #2608, opened 2021-05-14
fix(close): Header example with button close

ajaypatelaj:
BS5 changed the close button class to btn-close: https://getbootstrap.com/docs/5.0/components/close-button/
Added class with float utility

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -1135,7 +1135,7 @@
       if (this.options.header) {
         header =
           '<div class="' + classNames.POPOVERHEADER + '">' +
-            '<button type="button" class="close" aria-hidden="true">&times;</button>' +
+            '<button type="button" class="btn-close close float-end" aria-hidden="true"></button>' +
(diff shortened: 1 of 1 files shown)


caseyjhol:
This breaks the close button in Bootstrap 3 and Bootstrap 4. A solution may be to:
- remove &times; from inside the button
- via CSS, set `.close` content to the unicode for &times;
- set `classNames.CLOSEBUTTON = 'close'` by default. If version is >= 5, set `classNames.CLOSEBUTTON = 'btn-close float-end'`

caseyjhol (changes requested):
This breaks the close button in Bootstrap 3 and Bootstrap 4. A solution may be to:
- remove &times; from inside the button
- via CSS, set `.close` content to the unicode for &times;
- set `classNames.CLOSEBUTTON = 'close'` by default. If version is >= 5, set `classNames.CLOSEBUTTON = 'btn-close float-end'`

PR #2621, opened 2021-06-10
Update defaults-tr_TR.js

mekutluca:
Updated 'deselectAllText' translation to make sure it conveys right message.

Previously, it was 'Seçiniz' which means 'Select', and it was a wrong translation. But now it says 'Tüm Seçimleri Kaldır' which means 'Deselect All' and it is correct now.


NicolasCARPi:
Thank you! @caseyjhol should be merged!

NicolasCARPi (approved):

PR #2634, opened 2021-07-13
Remove IE9 hack causing issues in 2021. Fix #1122

### LINKED ISSUE #1122: Include the: "width: 220px" on the: "bootstrap-select.css" for the non Internet Explorer Browsers.
When using bootstrap-select in Chrome, it detect the property: "width: 220px \0;" as invalid, and according to:
http://stackoverflow.com/questions/8004765/css-9-in-width-property
it should be defined in the "bootstrap-select.css" file also another with property for the others browsers, something like:

``` javascript
.bootstrap-select {
  width: 220px \0;
  /*IE9 and below*/
  width: 220px;
}
```

NicolasCARPi:
Note: https://github.com/snapappointments/bootstrap-select/issues/2617 should really be adressed!


NicolasCARPi:
fixed by https://github.com/snapappointments/bootstrap-select/commit/17cffd46760365bcb4a590d0efb445d3b48a02dd

PR #2650, opened 2021-08-26
Fix long text option overflow

fathilarham:
Add max-width 100% in .dropdown-menu class to prevent text overflow.


NicolasCARPi (approved):

NicolasCARPi:
Hello,

Can you show with examples (and screenshots) what it fixes? Also, I'm not sure it's the good approach. `overflow` might be a better css approach to this issue.

fathilarham:
Hi @NicolasCARPi, 
The max-width will fix the dropdown options overflow if we put that in the bootstrap col class.

Without max-width:
image

With max-width:
image

PR #2660, opened 2021-09-27
Add support for tags

adaniello:
Converts a select with multiple attribute and `tags` option set to true, in a tags input: by searching for tags, if tag isn't in list user can add by hit Enter.

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -840,9 +840,14 @@
 
   function showNoResults (searchMatch, searchValue) {
     if (!searchMatch.length) {
-      elementTemplates.noResults.innerHTML = this.options.noneResultsText.replace('{0}', '"' + htmlEscape(searchValue) + '"');
-      this.$menuInner[0].firstChild.appendChild(elementTemplates.noResults);
+      this.no_results = true;
+      if (!this.options.tags) {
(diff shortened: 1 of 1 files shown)


NicolasCARPi:
Hello,

Thank you for your proposition. I think at the moment we need to focus on fixing existing bugs and releasing a new version, so don't be surprised if this PR, which adds code/feature is not reviewed quickly. Is there an issue linked to this change?

adaniello:
I agree. Simply this feature was requested many times: see at https://github.com/snapappointments/bootstrap-select/issues/565 and https://github.com/snapappointments/bootstrap-select/issues/2228#issuecomment-508869398

Thanks

akinuri:
Is there a demo/example page for this?

I tried this on the [Examples](https://developer.snapappointments.com/bootstrap-select/examples/) page, but it didn't work.

* Selected the `select` element in the "Live search" example
* Ran `$0.dataset.tags = "true"`
* Ran `jQuery($0).selectpicker("refresh");`
* Searched for a nonexistent value and hit Enter

Nothing happens.

<hr>

I just wrote something similar (but fancier) to this, and wanted to know if the plugin already/natively supports this behaviour. Came across this, but this does not seem to be documented.

Organizer21:
+1 for this being a long overdue feature.

PR #2661, opened 2021-09-27
Update Carbon ad code

sayzlim:
Replacing the legacy placement based identifier with zone key identifier. You can check the latest ad code by signing into sell.buysellads.com with your email address.

### DIFF
--- a/docs/custom_theme/main.html
+++ b/docs/custom_theme/main.html
@@ -62,7 +62,7 @@
   <div class="container-fluid">
     <div class="carbonad">
       <div class="carbonad-inner">
-        <script async type="text/javascript" src="//cdn.carbonads.com/carbon.js?zoneid=1673&serve=C6AILKT&placement=silviomoretogithubiobootstrapsel" id="_carbonads_js"></script>
+        <script async type="text/javascript" src="//cdn.carbonads.com/carbon.js?serve=CKYIKK7I&placement=developersnapappointmentscom" id="_carbonads_js"></script>
       </div>
     </div>
   </div>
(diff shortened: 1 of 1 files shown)


NicolasCARPi:
Hello,

I see that both files are the same and you are the dev of carbonads so nothing fishy here ;)

Can you also add the sha integrity to the tags?

sayzlim:
@NicolasCARPi let me check with the developer of the team. I'm only responsible for communicating the changes to the site owners. Will keep you updated.

Just want to be sure, do you mean something like this: [Subresource Integrity - Web security](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity)

I'm afraid that we don't use `integrity` value because you will have to update them whenever we make change to the file. Is it okay if we're using the ad code with the current structure?

NicolasCARPi:
> I'm afraid that we don't use integrity value because you will have to update them whenever we make change to the file.

Yes that's how it works. Now the question I have is: how often is this code changed for a good reason? When loading things from a CDN, SRI can save users from being infected by malware if the CDN gets compromised.

sayzlim:
We will only update it when we discover a bug with the script, otherwise we won't make any changes unless necessary. I'm still going to check with the developer whether it makes sense for us to enforce SRI.

NicolasCARPi:
It is true that it makes more sense for versioned assets. Where version X.Y.Z of a lib must always have the same hash. But for a script serving ads, maybe not.

sayzlim:
@NicolasCARPi appreciate the feedback. We've been running the same script for other open-source projects for years, and we haven't seen any needs to enforce SRI. Feel free to accept this PR if you're okay with it, or I can get back to you when I have more information from our developers.

NicolasCARPi (approved):

sayzlim:
@NicolasCARPi seems the site is not updated yet. Looks like the site is the mirror of **gh-pages** branch, should I make an update on the branch and submit the PR again?

NicolasCARPi:
@sayzlim I'm not sure, that would be a question for @caseyjhol because I don't know his setup.

caseyjhol:
We're currently hosting the docs separately, but we're hoping to switch to gh-pages in the near future to make it easier to update.

PR #2669, opened 2021-10-13
Replace division with multiplication

senchden:
Addresses https://github.com/snapappointments/bootstrap-select/issues/2668. Removed slash character to be compatible with the newer versions of `sass` compiler. Tested and does not brake compatibility since resulting CSS is not changed.

### DIFF
--- a/sass/bootstrap-select.scss
+++ b/sass/bootstrap-select.scss
@@ -23,7 +23,7 @@
 
 @function fade($color, $amnt) {
   @if $amnt > 1 {
-    $amnt: $amnt / 100; // convert to percentage if int
+    $amnt: $amnt * 0.01; // convert to percentage if int
   }
   @return rgba($color, $amnt);
 }


NicolasCARPi:
Thanks, indeed this should not cause any issues. But the [sass doc](https://sass-lang.com/documentation/breaking-changes/slash-div) recommends using `math.div`. I think it is better to use `math.div` as it might handle better different values of `$amnt` (compared to the `*` operator). But this is pure supposition, maybe the check if `> 1` above is enough to make sure no weird values are there. Anyway, I would still side with the documented change.

senchden:
Yes, it could be replaced with `math.div`. But I have done it this way if to keep backwards compatibility with `libsass`/`node-sass` (just like in bootstrap). In fact, multiplication is what official sass migrator came up with. If you don't need it, I can replace it with `math.div`

NicolasCARPi:
For me you provided a valid reason to use the multiplication so let's use that then! Thanks!

NicolasCARPi:
dammit, I fat fingered the commit message.... oh well....

PR #2684, opened 2021-11-27
Disabled btn-light

jiriermis:
Better compatibility with bootstrap-dark. With disabled btn-light look much better in bootstrap-dark mode. Even in light mode look still very good.

Make sure to target the `dev` branch!


NicolasCARPi:
No, that's not how it should be done. Also, read the contributing doc.

PR #2689, opened 2021-12-07
fix issue with title attribute used to display alternative text

### LINKED ISSUE #2674: Using title on options for alternative text doesn't work
https://developer.snapappointments.com/bootstrap-select/examples/#selected-text
Don't work for 1.14beta version

NicolasCARPi:
fix #2674


fabienwnklr (approved):
Good job !

Issue #2697, opened 2022-01-12
Bootstrap-select v1.14.0-beta2 dropdown menu is not hiding when we press enter after selecting any option.


PeterGrand:
I found this to be the case with Bootstrap 5.1.3. The widget would work, but when you start typing to select an option and then press enter to set it, it would set the value but not close the drop down. We wrote a quick patch for the issue to get us back up and running. Will create a PR now.

### Discussed in https://github.com/snapappointments/bootstrap-select/discussions/2638

<div type='discussions-op-text'>

<sup>Originally posted by **Adityageek** July 28, 2021</sup>
Hi guys, currently I am migrating to Bootstrap-select v1.14.0-beta2 with bootstrap 5 but the dropdown menu is not hiding when we press Enter key after selecting an option, but the same is working with the click event, anyone knows the issue. Please help me with this issue.
 </div>

dmalan:
Hi @PeterGrand, just wanted to ask if there are plans to release an `alpha4` that would include https://github.com/snapappointments/bootstrap-select/pull/2698 (and propagate to, e.g., https://www.jsdelivr.com/package/npm/bootstrap-select)?

Thanks for the terrific library!

PR #2698, opened 2022-01-12
Close drop down when selecting item with enter key

PeterGrand:
This ensures that the drop down closes after choosing an item by pressing the enter key.

Fixes [#2697](https://github.com/snapappointments/bootstrap-select/issues/2697)

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -3175,8 +3175,10 @@
             $(document).data('spaceSelect', true);
           }
         }
-        // hide dropdown menu
-        that.dropdown.hide();
+        if (e.which === keyCodes.ENTER) {
+          // hide dropdown menu
+          that.dropdown.hide();
+        }
       }
     },
 
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -3175,6 +3175,8 @@
             $(document).data('spaceSelect', true);
           }
         }
+        // hide dropdown menu
+        that.dropdown.hide();
       }
     },
 


NicolasCARPi:
Hello,

Thank you for proposing a PR to fix an issue.

I cannot reproduce the bug in question: in my hands, pressing enter closes the dropdown. Tried with chromium and firefox on `tests/bootstrap5.html`. Is the bug reproducible?

PeterGrand:
I'll see if I can figure out the trick to reproducing. I saw two people mentioning it in discussions and we found it when we updating our django app to Bootstrap 5 with this package.

NicolasCARPi:
I don't mind merging it as it's a rather small change, but I also know that the smaller the change, the higher risk it has to break other stuff, so that's why I would prefer if it fixes a reproducible issue.

SGunet:
If I may suggest 3 things : 

- I would close the dropdown on Enter or Tab if selectOnTab option is activated
- I would not call e.preventDefault() to let the focus jump to the next control when you validate your selection
- It might also seem more readable to insert your code on line 3170, just under this.trigger('focus'), which is called right after the item is actually selected.

Sorry if this is not the right way to comment on somebody else's pull request.

derklaro:
Hi, I don't want to unessacarily bump this but I can confirm that this issue is still present in alpha3 (we're currently in the upgrade process to bootstrap 5 as well). I wasn't able to track down any issues on our end why the dropdown menu isn't closing, but maybe I oversaw something...

NicolasCARPi:
@derklaro And does this change fixes the issue you see?

derklaro:
It does! :)

NicolasCARPi (approved):

jp41011:
Looks like this pull request updated the **"js/bootstrap-select.js"** file but not the **"js/bootstrap-select.min.js"** file.
Just FYI for anyone that might be trying to use this fix, the **min** and **map** files were not updated with the changes included in the **"js/bootstrap-select.js** file.

PR #2710, opened 2022-02-16
Fix ajax search and maxOptions. Add initial Cypress tests

### LINKED ISSUE #2593: maxOptions display bug
jQuery version: 3.5.1
Bootstrap version: 5.0.0-beta3
bootstrap-select version: 1.14.0-beta2
Browser: Firefox

If I create a select with BS5 which is limited by the "max-options" parameter and click on options after the max limit is reached, the options will be still marked with an icon. But these options aren't really selected, so it seems it is an display bug.

Here is a codepen example: https://codepen.io/dernst/pen/abpperx

### LINKED ISSUE #2659: Ajax search get multiple calls
Hi,
i'm trying to use Ajax/JSON data source support as in https://github.com/snapappointments/bootstrap-select/releases/tag/v1.14.0-beta.

No problem if i use `source.data` property with an ajax call, but when i try to use `source.search` i get a very strange issue: typing 1 character, selectpicker posts only one ajax call; typing 2 characters, selectpicker posts 2 ajax call; from 3 characters posts so much ajax call, lower is result greather is number of calls (eg. with only two items in result, selectpicker post also 21 ajax call repeating more time result).

This is my JS code:
```
$(function () {
    $('#test').selectpicker({
        source: {
            data: function (callback) {
                $.ajax({
                    method: 'POST',
                    url: 'test_selectpicker.php',
                    dataType: 'json'
                }).then((response) => callback(response.data))
            },
            search: function (callback, page, searchTerm) {
                let data = { page, search: searchTerm };
                $.ajax({
                    method: 'POST',
                    url: 'test_selectpicker.php', 
                    data,
                    dataType: 'json'
                }).then((response) => callback(response.data))
            }
        }
    });
});
```

Thanks

### LINKED ISSUE #2671: Ajax search not works
Ajax/JSON support, "the highlight" of 1.14.0-beta release (https://github.com/snapappointments/bootstrap-select/releases/tag/v1.14.0-beta), seems not works.

My code is very simple (see at https://codepen.io/adaniello/pen/xxLVrKE) and returns some issues:
1. ajax search, in some conditions, renders a list with some items duplicated;
2. ajax search with two results post an high number of ajax calls due a logical error (as reported in https://github.com/snapappointments/bootstrap-select/issues/2659);
3. "defocus", eg. by typing arrows, not works;
4. selection not works.

Is there someone who managed to make it work? 

I already fixed 1 and 2 points (a pull-request will be available when all point fixed), but i'ld like to get some feedback on this issues.

Thanks

caseyjhol:
- fixes #2659
- fixes #2671
- fixes #2593
- add some basic Cypress tests/workflow
- support passing in a boolean `more` argument as an additional argument to the callback function
- remove `source.load` method in favor of simply using `source.data`. `page` is now passed into `source.data` as the 2nd argument (just like `source.search`). If you want static data on init (and don't want to call your AJAX function), you can just check `page === 0`).

Demo: https://plnkr.co/edit/1j2mceV5ewKIxp9n


caseyjhol:
@NicolasCARPi This is still a WIP. I'll send it your way to review once I'm done.

js/bootstrap-select.js
```diff
@@ -1021,15 +1024,6 @@
         });
       });
 
-      this.fetchData(function () {
```
caseyjhol, line 1024: removing this duplicate block

caseyjhol:
@NicolasCARPi I'm still working on fixes for #2671, but there's no reason those should hold up this fix. Let me know your thoughts on dropping the `source.load` method.

NicolasCARPi (changes requested):
I think it would be best to have a test case in the `tests` folder, with a fake ajax call (just a function returning stuff), so we can actually assess if the fix works and have different test case (you said the bug wasn't found before because it was only triggered in certain conditions).

We don't have automated tests (yet), but I believe it's important to add test cases when possible.

caseyjhol:
@NicolasCARPi Have you made any progress on implementing Cypress? I also have Cypress experience and might try to tackle some of the initial groundwork if you haven't already.

NicolasCARPi:
Nope, I didn't, so feel free to lay the groundwork ;)

NicolasCARPi (approved):
I think it's great to finally have some end to end testing! This should be merged and a new beta version must be released ASAP :)

caseyjhol:
@NicolasCARPi Yes it's about time. I'm excited to add more. Going to make it so much easier to add new features and fix bugs. Just need to update the changelog and I think we can prep for a release. I'd like to also look into it to see if there's a way to automatically create a nuget release at the same time as well.

NicolasCARPi:
> create a nuget release

sure, see: https://github.com/marketplace/actions/publish-nuget

Keep in mind that github actions have issues right now ;) https://www.githubstatus.com/

PR #2713, opened 2022-02-28
[#2667]fix the bug of the data-size when the first option style has 'display: none'

### LINKED ISSUE #2667: Hide elements with data-size attribute
bug with `data-size`
there is a large list of elements with a search string, when I add `data-size`, all the elements disappear, `max-width` is added to some elements in the code and everything disappears.
all the presented elements are added via js

---

megumiimai:
resolved: #2667

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -1454,7 +1454,7 @@
     buildData: function (data, type) {
       var dataGetter = data === false ? getOptionData.fromOption : getOptionData.fromDataSource;
 
-      var optionSelector = ':not([hidden]):not([data-hidden="true"])',
+      var optionSelector = ':not([hidden]):not([data-hidden="true"]):not([style*="display: none"])',
           mainData = [],
           startLen = 0,
           optID = 0,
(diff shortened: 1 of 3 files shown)


js/bootstrap-select.js
```diff
@@ -1920,7 +1920,8 @@
       if (this.selectpicker.current.data.length) {
         for (var i = 0; i < this.selectpicker.current.data.length; i++) {
           var data = this.selectpicker.current.data[i];
-          if (data.type === 'option') {
+          var a = data.element.firstChild;
```
NicolasCARPi: `a` is a poor variable name choice

NicolasCARPi (changes requested):
Thank you for your contribution. This change would be a good opportunity to add a test case in the `tests` folder (even if we don't currently test automatically for things yet). Basically an example showing the problem, and showing that with this PR the problem goes away!

megumiimai:
thank you for reviewing. I added a test (febdfea)  and pasted the capture. How about that?

### problem


### fixed

NicolasCARPi (approved):
Thanks, looks good to me!

PR #2767, opened 2022-08-02
Fix dropdown position when use data-size attribute

### LINKED ISSUE #2766: When data-size attribute is on, isDropup flag got wrong value
When select has a lot of option (in my case 100+) I wanted to limit it to display only the 6 options with scroll. I use attribute data-size set to 6. But when i click to open select, isDropup flag is set to true, and option show above a select but is enough space to display options below button.

probably problem is in line:
estimate = liHeight * this.selectpicker.current.elements.length + menuPadding.vert;

proposition of fix that looks like it's working:
if (this.options.size !== 'auto') {
          estimate = liHeight * this.options.size + menuPadding.vert;
        } else {
          estimate = liHeight * this.selectpicker.current.elements.length + menuPadding.vert;
        }

bootstrap-select ver. 1.13.14

smiglu:
When select has a lot of option (in my case 100+) I wanted to limit it to display only the 6 options with scroll. I use attribute data-size set to 6. But when i click to open select, isDropup flag is set to true, and option show above a select but is enough space to display options below button.

Fix #2766

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -2250,11 +2250,8 @@
         // This is useful for smaller menus, where there might be plenty of room
         // below the button without setting dropup, but we can't know
         // the exact height of the menu until createView is called later
-        if (this.options.size !== 'auto') {
-          estimate = liHeight * this.options.size + menuPadding.vert;
-        } else {
-          estimate = liHeight * this.selectpicker.current.data.length + menuPadding.vert;
-        }
+        var dropDownLength = this.options.size !== 'auto' ? this.options.size : this.selectpicker.current.data.length;
+        estimate = liHeight * dropDownLength + menuPadding.vert;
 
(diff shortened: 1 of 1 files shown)


js/bootstrap-select.js
```diff
@@ -2250,7 +2250,11 @@
         // This is useful for smaller menus, where there might be plenty of room
         // below the button without setting dropup, but we can't know
         // the exact height of the menu until createView is called later
-        estimate = liHeight * this.selectpicker.current.data.length + menuPadding.vert;
+        if (this.options.size !== 'auto') {
```
NicolasCARPi: you could avoid an `else` with a var holding either `this.options.size` or `this.selectpicker.current.data.length` depending on the value of `this.options.size`. And it would reduce code duplication for the `estimate` var.

smiglu: ok

NicolasCARPi:
I can't reproduce the issue you describe (with select appearing as dropup). Could you try and have a reproducible case in `tests` that shows that your fix is useful?

PR #2816, opened 2023-02-07
fix refresh action

SimonTod:
The refresh function is broken.

It duplicates the lines in the dropdown as well as the button title text.

How to reproduce :

- checkout main or dev branch.
- open tests/bootstrap5.html
- select a value with a selectpicker
- trigger the refresh function on the modified select (using browser console)

As a result :

- button title is duplicated
- drowdown elements are duplicated

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -3393,6 +3393,8 @@
       var config = $.extend({}, this.options, getAttributesObject(this.$element), this.$element.data()); // in this order on refresh, as user may change attributes on select, and options object is not passed on refresh
       this.options = config;
 
+      this.selectpicker.main.data = [];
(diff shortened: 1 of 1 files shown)


nftstudio:
Nice,it worked!

cwolcott:
Fixes the basic issue for v1.14.0-beta3.

tiesont:
@NicolasCARPi Any idea of when a new release will be pushed out? This isn't in any CDN version of which I'm aware

NicolasCARPi:
@tiesont never. This project is completely dead and must not be used. We moved to Tom Select. Unfortunately, author didn't make it extra clear that this project is unmaintained. It should be archived.

PR #2844, opened 2023-09-02
[#2826]trim values before search

### LINKED ISSUE #2826: Trim values before search in selectpicker
#### Env: Window-11, Wifi connection, chrome-107v
----------------
1. Open the select picker 
2. Input vales in search box which have results 
3. Now take the cursor to the starting of text 
4. now input space 
5. observe the result does not match the text we input 
----------------------------------------------
Expected : the result should visible as per the text we input in search 
----------------------------
ezgif com-video-to-gif

megumiimai:
resolved #2826 

56d3809900af8d017439eda886ca7a66

Make sure to target the `dev` branch!

### DIFF
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -2953,6 +2953,7 @@
         var searchValue = that.$searchbox[0].value;
         var isWhitespace = /^\s*$/.test(searchValue);
         if (!isWhitespace) {
+          // trim leading and trailing half-width spaces and full-width spaces.
           searchValue = searchValue.replace(/^\s+|\s+$/g, '');
         }
 
--- a/js/bootstrap-select.js
+++ b/js/bootstrap-select.js
@@ -2951,6 +2951,10 @@
 
       this.$searchbox.on('input propertychange', function () {
         var searchValue = that.$searchbox[0].value;
+        var isWhitespace = /^\s*$/.test(searchValue);
+        if (!isWhitespace) {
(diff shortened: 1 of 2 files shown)


js/bootstrap-select.js
```diff
@@ -2951,6 +2951,10 @@
 
       this.$searchbox.on('input propertychange', function () {
         var searchValue = that.$searchbox[0].value;
+        var isWhitespace = /^\s*$/.test(searchValue);
+        if (!isWhitespace) {
+          searchValue = searchValue.replace(/^\s+|\s+$/g, '');
```
caseyjhol, line 2957: I think we're probably okay to just use `searchValue = searchValue.trim();` instead.

caseyjhol (changes requested):

js/bootstrap-select.js
megumiimai: @caseyjhol 
thank you very much.
Since some users sometimes enter full-width spaces, so I have added support for trimming all half-width spaces and full-width spaces.

caseyjhol: Ah gotcha. Maybe just mention this in a brief comment above this line.

caseyjhol (changes requested):

js/bootstrap-select.js
megumiimai: I have added a comment. thank you.

caseyjhol (approved):

